PR #2, opened 2015-06-06
Test and advertise Python 3 and PyPy support

jmbowman:
I've started tracking the Python dependencies of my packages using [requires.io](https://requires.io/) and noticed that although pytest-catchlog works with Python 3, that site doesn't list it as being Python 3 compatible.  Same for [caniusepython3.com](https://caniusepython3.com/project/pytest-catchlog).  As pointed out in the [caniusepython3 README](https://github.com/brettcannon/caniusepython3#how-do-you-tell-if-a-project-has-been-ported-to-python-3), this is because setup.py doesn't include a `Programming Language :: Python :: 3` trove classifier.  I've added classifiers for the versions I've personally verified and the config file for [tox](https://tox.readthedocs.org/en/latest/) that I used to test them.


eisensheng:
Version 1.1 is online now. Thanks for contributing!

PR #6, opened 2015-09-06
Add support for setting fallback values for options through ini file

abusalimov:
That is, `--log-format` option maps to `log_format` setting within `[pytest]` section of pytest.ini / tox.ini / setup.cfg, and `--log-date-format` - to `log_date_format`.

Now, instead of writing:

```
[pytest]
addopts = --log-format='%(message)s ...'
```

One can set the `log_format` directly:

```
[pytest]
log_format = %(message)s ...
```

The advantage of the latter is that the `addopts` approach forces end users to have pytest-catchlog installed, because otherwise pytest invocation fails with `unrecognized arguments` error. In case of using `log_format = ...` everything works fine as if there was no such option.

I believe, such behavior should be implemented and handled by pytest itself, but still...

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -58,6 +58,18 @@ Shows failed tests like so::
     text going to stderr
     ==================== 2 failed in 0.02 seconds =====================
 
+These options can also be customized through a configuration file::
+
+    [pytest]
+    log_format = %(asctime)s %(levelname)s %(message)s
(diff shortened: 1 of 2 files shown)


eisensheng:
Thank you very much for you contribution! The proposed Feature and Changeset is beautiful.

abusalimov:
Thank you for taking the maintenance burden of this cool plugin!

Issue #3, opened 2015-07-01
__multicall__ being deprecated in upcoming pytest release


The-Compiler:
When using pytest-catchlog with pytest from [their git repo](https://github.com/pytest-dev/pytest) and `-rw` (show warnings), I get:

```
.../pytest_catchlog.py:71 'pytest_runtest_makereport' hook uses deprecated __multicall__ argument
```

The-Compiler:
I'm thinking about fixing this separately from #7 as I really want to switch to pytest 2.8 without getting warnings :wink:

@eisensheng Would you do a new release for me if I fixed this?
@abusalimov Is it okay if I introduce merge conflicts in your PR (sorry!)?

abusalimov:
@The-Compiler Sorry for delaying this show stopper.

I've stripped out any API-breaking changes from the PR recently, but have not pushed yet because of few failing tests.
I arrived to Prague yesterday and now I have better Internet here and a bit more quite time. I'm going to fix remaining tests and force-push #7 this evening so that you could review the corresponding `__multicall__` fix.

Is it OK for you?

The-Compiler:
Ah, awesome! Sounds good to me.

Probably makes sense to move the API/LineMatcher changes to a separate PR anyways as they might take a little bit more time/discussion.

Thanks for all your work on catchlog, I really appreciate it!

eisensheng:
> I've stripped out any API-breaking changes from the PR recently, but have not pushed yet because of few failing tests.

Too bad. I was all in favour for the API break. We could simply release 1.2 with the feature to pull configuration from the cfg file and a fix for `__multicall__` and introduce the API break with a 2.0 version.

This would also be the safest upgrade path for current users of the original plugin and this fork. (If there are any users of this fork that is …)

abusalimov:
That doesn't mean that I've thrown all that changes into to a trashcan :recycle:
I was going to suggest them in a separate PR, just not to block #7 for too long.

Issue #1, opened 2015-04-22
Option to always report logs


The-Compiler:
Hi!

In 19ff0645a910a0a7d545e7e0baf7eed350ddb93c you mention:

> A far better approach would be to replace this option with an
> 
> ```
>     --report-logs=[never,failed,always]
> ```
> 
> option though.

For my project, I use a buildbot which marks the tests as "warning" when there was logging output. I also use debug logging a lot, so ideally I'd be able to do something like this:
- Report log entries with debug/info level when a test failed.
- Always report log entries with warning/error/critical levels.

If something like that would be a welcome addition, please let me know and I'll submit a pull request!

Inputs on how the configuration would look would be welcome though. Maybe keep `--no-print-logs` and then add `--loglevel-pass` and `--loglevel-fail` or something?

eisensheng:
Hi!

Thanks for your input. My memory on this project is a little cloudy. IIRC the primary need for me to work on this library was the requirement to ensure that the application logs certain events and thus the log lines had to be caught. Now you're proposing a second use case for this library to catch any log lines and output logs for failing tests. 

I like the idea since it seems to me to be very useful and thus would love to see a patch for it and also would merge it if you would be so kind to provide one. I would suggest a different approach to it though to keep the explanation of the feature simple. I'll try to collect some thoughts on this here:
- I think it is always a desired behavior to receive all logged events for a failed test IMHO. Following that thought means there is no need to configure a minimum level for failed tests. 
- It would also be nice to still be able to silence the log output altogether so you won't litter the logs on your CI system.
- It would still be nice to see logs of a certain level on your local development system to spot possible errors. 
- To enable the behavior on a global scope that any log output may lead to a non successful state of the test seems not that good to me TBH. It may even be the case that an error event here and a warning event there is totally fine because you're testing a fallback behavior for a failed condition. If this scenario stays valid it means that a global configuration for this may not fit the need of all your tests. The need to fail tests if an error/warning/critical event was caught can be realized through an assert though.

It seems to me to be quite the difficult endeavor and there is further need to iron out the remaining wrinkles in the proposal. I hope the text was understandable and constructive to you and didn't kill your motivation regarding this issue. It was a tiresome workday for me so I might have overlooked something. I would thank you if you could stay patient with me and point it out to me then. Thanks!

The-Compiler:
I agree with the first three points! So the only thing needed from an UI perspective would be some option to set a minimum loglevel to always show, right?

[...]

PR #7, opened 2015-09-08
Capture logs also from setup and teardown stages

abusalimov:
Now for failing tests that log anything during setup/teardown, the corresponding output is reported in separate sections.

Example:

```
------------------------ Captured stdlog setup ------------------------
DEBUG      ...
------------------------ Captured stdlog call -------------------------
INFO       ...
```

Issues and drawbacks:
- ~~Outputs weird _Captured **std**log ..._ captions - should be better fixed it pytest, I think~~ (**UPD**: addressed in pytest-dev/pytest#1052)
- Requires recent versions of pytest, at least 2.6 - one and a half year old at the moment of suggesting the PR (the latest one is 2.7.2 now)

Please let me know if you believe these issues must be addressed before merging.


abusalimov:
It's a bit late at my home now, I'll try to fix test failures later this week.

eisensheng:
I reviewed your work so far and I really like what you did so far! It also appears that merging this change set will also fix #3 since you dropped the code that relied on the deprecated `__multicall__` interface. All the more reason to merge this!

> Issues and drawbacks:
> - Outputs weird Captured stdlog ... captions - should be better fixed it pytest, I think

Yep, it looks like this is caused by https://github.com/pytest-dev/pytest/blob/2093889ac2f346c232ca86540a9802de58cb6f5b/_pytest/runner.py#L229. I would say that this wart shouldn't prevent me from merging and releasing it though. I would say it's their problem and their responsibility to fix it.

> - Requires recent versions of pytest, at least 2.6 - one and a half year old at the moment of suggesting the PR (the latest one is 2.7.2 now)

It looks to me that the merge request also contains larger changes to the API of this plugin. I would say that this API change alone justifies a 2.0 release. So while we are at it we can also pull up the requirements for the dependencies in a new major version.

It appears to me that the Merge Request still lacks a few tests though regarding the original purpose of the request: Capturing logs from setup and tear down stages. It would be nice to have this functionality covered by tests.

Please tell me what you think about it. Thanks!

abusalimov:
Hi,

Thanks for reviewing and sorry for lots for "fixup" and "wip" commits, the
changeset still needs to be rebased and polished.

I'm on vacation at the moment, so I'm not able to finish the work right
now, but I'm definitely going to do that once I get some spare time and
internet access.

> It appears to me that the Merge Request still lacks a few tests though
> regarding the original purpose of the request

Yep, it also lacks proper documentation for new features like capturing
logs of a certain code block, or using methods inherited from
`pytester.LineMatcher`, e.g.

```
with caplog.recording() as rec:
    ...
    rec.fnmatch_lines(...)
```

Please do not merge until I fix that. I'll let you know, when everything
becomes ready.

Regards,
Eldar

pytest_catchlog.py
```diff
@@ -162,59 +190,96 @@ def record_tuples(self):
...
+    @property
+    def lines(self):
+        return [r.getMessage() for r in self.buffer]
 
-        By default, the level is set on the handler used to capture
-        logs. Specify a logger name to instead set the level of any
-        logger.
+    def __init__(self, handler):
+        # Do not call LineMatcher.__init__(),
+        # it attempts to assign to self.lines, which is a property.
```
The-Compiler: I wonder if inheriting from `LineMatcher` is a good idea in general - it might change in future pytest versions in ways which might break this - why not just copy `fnmatch_lines` from there and remove the magic we don't need here (like `_getlines`)?

The-Compiler:
I submitted a PR for pytest to remove the hardcoded `std`:

https://github.com/pytest-dev/pytest/pull/1052

eisensheng:
> I submitted a PR for pytest to remove the hardcoded std:

Very nice! Thanks for the fix. This will make it more appealing from an aesthetic point of view.

pytest_catchlog.py
eisensheng: I think it depends on whenever `LineMatcher` has been declared as public API whenever it should be reused or not. Duplicating code is something I'm not very fond of. And since it actually has been [documented](http://pytest.org/latest/writing_plugins.html?highlight=linematcher#_pytest.pytester.LineMatcher) I would say it's okaish unless you still would say it's still a big no. Your words have more weight compared to mine in this field since you're the one with more experience in py.test development.

The-Compiler: Good point - I didn't notice it was actually documented. I still think it's probably intended for internal pytest use and the documentation is intended for hooks returning a `LineMatcher` - but hey, what can possibly go wrong.

On a second thought, I'd say let's keep this and see if it breaks - which, after all, is unlikely.

The-Compiler:
The pytest PR will probably be moved to the `features` branch, i.e. pytest 2.9, because `Item.add_report_sect` was undocumented before. This means it'll probably take a while until the change lands.

I tried this branch with my project, and with some adjustments to my [logfail.py](https://github.com/The-Compiler/qutebrowser/blob/catchlog/tests/helpers/logfail.py) which fails tests on logging messages if `at_level` wasn't used, everything seems to work fine!

I definitely prefer the newer api (`records()` always bugged me...), but I wonder if it'd be a good idea to keep backwards compatibility. Many people are using `pytest-capturelog` at the moment, and at some point that'll likely die (because it's still unmaintained and `__multicall__` will probably be removed in a future pytest release). With this change, it'll be a lot harder to migrate from there...

I don't think it'll take much - `mixedCase` aliases for the `snake_case` methods, and `records` could return a `collections.Iterable` which returns `self` in `__call__`. But of course there's a future maintenance cost too.

pytest_catchlog.py
```diff
@@ -162,59 +190,96 @@ def record_tuples(self):
...
+    def recording(cls, **kwargs):
+        with catching_logs(RecordingHandler(), **kwargs) as handler:
+            yield cls(handler)
+
+    def at_level(self, level):
+        """Context manager that sets the level for recording of logs.
+
+        This only affects logs collected by this funcarg. To control
+        captured logs that are reported for failed tests use the
+        ``logging_at_level()`` context manager.
```
The-Compiler: What's the rationale for dropping the `logger`/`handler` argument? Just a cleaner API?

pytest_catchlog.py
abusalimov: My thought on reusing `LineMatcher`:

Pros.
- It is (widely?) used by plugins for self-testing (catchlog itself, for example); the API is unlikely to change
- Even if the API changes, that won't affect the main functionality provided by catchlog
- The logic provided by `LineMatcher` is somewhat hardly related to catchlog to copy-paste the code
- Reusing code is generally a better option

Cons.
- The code of `LineMatcher` is a bit dirty, IMO; the API could be more friendly for subclasses
- It _is_ a dependency, yet not explicit

My opinion: it would be cool to have line matching feature in the core capturing module of pytest (not pytester), i.e. to be able to make assertions about captured stdout/stderr as well.

Let's throw away this feature for now (at least from this PR), and try to think about that in more wide context: (maybe?) move `LineMatcher` from `pytester` to the `capture` plugin, realize what we want to have there, provide backward compatibility, etc.

Anyway, this again is a pytest issue / feature request. But once we have that feature for `capsys`, there would be no doubts to provide the same API for `caplog`.

abusalimov:
> I definitely prefer the newer api (`records()` always bugged me...), but I wonder if it'd be a good idea to keep backwards compatibility.

That comes from my WIP commit, so that's definitely not a final solution. Thank's for the good point anyway. Like as you suggested, I plan to subclass the `list` and add `__call__` emitting a deprecation warning and returning `self`. May there'd be also a good idea to provide something like `--capturelog-compat` flag for suppressing these warnings for case when there are lots of them printed, spamming the output.

> I don't think it'll take much - `mixedCase` aliases for the `snake_case` methods

Not a problem too.

> I tried this branch with my project, and with some adjustments to my logfail.py which fails tests on logging messages if `at_level` wasn't used, everything seems to work fine!

Looks nice! I was thinking about such feature, i.e. add a `--log-fail-on=ERROR` option. However, this part of your code is not so obvious to me:

```
        # caplog.at_level(...) was used with the level of this message, i.e.
        # it was expected.
```

It seems that things are bit more complicated than I thought. And I've just discovered that there is a discussions on this topic in #1.

pytest_catchlog.py
abusalimov: That's mainly because there're separate handler instances used for reporting and inspection now. And naively changing a level of a logger through a `at_level` call on some recorder would affect all attached handlers, i.e. other recorders. This looks wrong to me, despite there used to be the same behavior before.

Honestly, I don't know what to do about that. It is possible to use a separate recorder for a certain `foo.bar` logger on some level, and even adjust the level in the middle of recording:

```
with caplog.recording(logging.INFO, logger='foo.bar') as rec1, \
     caplog.recording(logging.INFO, logger='some.other.module') as rec2:

    # Capture INFO and above.
    ...

    with rec1.at_level(logging.WARNING):
        # Capture only WARNING and below.
        # This doesn't affect a level of rec2 though.
        ...
```

But there's no way to change a level on, say, `foo.bar.baz` logger, except for `caplog.logging_at_level()`, which changes the level of the logger itself, affecting other recorders, including log reporting on failure.

I'll think about that. May be we could use log Filters for that purpose. I'd like to preserve backward compatibility after all.

The-Compiler:
> Like as you suggested, I plan to subclass the list and add **call** emitting a deprecation warning and returning self.

Note subclassing `list` isn't a good idea as subclassing a builtin comes with many fun surprises when C code uses `list` instead of the subclass internally.  Something like this would be better I think (untested):

``` python
import collections

class CallableList(collections.MutableSequence):

    def __getitem__(self, key):
        return self._data[key]

    def __setitem__(self, key, val):
        self._data[key] = val

    def __delitem__(self, key):
        del self._data[key]

    def __len__(self):
        return len(self._data)

    def insert(self, index, obj):
        self._data.insert(index, obj)

    def __call__(self):
        return self._data
```

there's also a more complete version as `collections.UserList`, but only in Python 3... here's the code (straight from the stdlib):

``` python
class UserList(MutableSequence):
    """A more or less complete user-defined wrapper around list objects."""
    def __init__(self, initlist=None):
        self.data = []
        if initlist is not None:
            # XXX should this accept an arbitrary sequence?
            if type(initlist) == type(self.data):
                self.data[:] = initlist
            elif isinstance(initlist, UserList):
                self.data[:] = initlist.data[:]
            else:
                self.data = list(initlist)
    def __repr__(self): return repr(self.data)
    def __lt__(self, other): return self.data <  self.__cast(other)
    def __le__(self, other): return self.data <= self.__cast(other)
    def __eq__(self, other): return self.data == self.__cast(other)
    def __ne__(self, other): return self.data != self.__cast(other)
    def __gt__(self, other): return self.data >  self.__cast(other)
    def __ge__(self, other): return self.data >= self.__cast(other)
    def __cast(self, other):
        return other.data if isinstance(other, UserList) else other
    def __contains__(self, item): return item in self.data
    def __len__(self): return len(self.data)
    def __getitem__(self, i): return self.data[i]
    def __setitem__(self, i, item): self.data[i] = item
    def __delitem__(self, i): del self.data[i]
    def __add__(self, other):
        if isinstance(other, UserList):
            return self.__class__(self.data + other.data)
        elif isinstance(other, type(self.data)):
            return self.__class__(self.data + other)
        return self.__class__(self.data + list(other))
    def __radd__(self, other):
        if isinstance(other, UserList):
            return self.__class__(other.data + self.data)
        elif isinstance(other, type(self.data)):
            return self.__class__(other + self.data)
        return self.__class__(list(other) + self.data)
    def __iadd__(self, other):
        if isinstance(other, UserList):
            self.data += other.data
        elif isinstance(other, type(self.data)):
            self.data += other
        else:
            self.data += list(other)
        return self
    def __mul__(self, n):
        return self.__class__(self.data*n)
    __rmul__ = __mul__
    def __imul__(self, n):
        self.data *= n
        return self
    def append(self, item): self.data.append(item)
    def insert(self, i, item): self.data.insert(i, item)
    def pop(self, i=-1): return self.data.pop(i)
    def remove(self, item): self.data.remove(item)
    def clear(self): self.data.clear()
    def copy(self): return self.__class__(self)
    def count(self, item): return self.data.count(item)
    def index(self, item, *args): return self.data.index(item, *args)
    def reverse(self): self.data.reverse()
    def sort(self, *args, **kwds): self.data.sort(*args, **kwds)
    def extend(self, other):
        if isinstance(other, UserList):
            self.data.extend(other.data)
        else:
            self.data.extend(other)
```

> Looks nice! I was thinking about such feature, i.e. add a --log-fail-on=ERROR option.

I'd be glad if something like this was built into `pytest-catchlog`. I was a bit stressed at the time I implemented it, and really didn't want to go through the pull-request/contribution dance (sorry!), so I did it separately.

> However, this part of your code is not so obvious to me: [...]

I have lots of tests which call functions in a "wrong" way, which causes them for example to log errors. In this case, I don't want to fail a test.

As these tests already use `at_level` (because the expected log message is at that level), I don't fail the tests in this case.

The-Compiler:
Another possibility for backwards compatibility would be to have a `caplog` and a `catchlog` fixture. Then users can clearly "opt-in" to the new API, and it's probably also cleaner to explain "caplog is deprecated, catchlog is the recommended thing but not backwards-compatible".

eisensheng:
I would rather suggest/prefer to solve this problem by simply adhering to semver and going straight to a 2.0 version with this feature and explicitly state the API break.

abusalimov:
@eisensheng I believe it's worth trying to preserve backward compatibility, at least for some time. @The-Compiler had a good point on this:

> Many people are using pytest-capturelog at the moment, and at some point that'll likely die (because it's still unmaintained and `__multicall__` will probably be removed in a future pytest release). With this change, it'll be a lot harder to migrate from there...

It is easier to pull down than to build, after all.

eisensheng:
> @eisensheng I believe it's worth trying to preserve backward compatibility, at least for some time.

That's also fine by me. I appreciate your enthusiasm. :+1:

abusalimov:
@eisensheng @The-Compiler Hey guys, I force-pushed the branch leaving out any changes besides the topic (including API-breaking changes) and reorganizing the changeset a little bit.
Also I didn't add `mixedCase`-to-`snake_case` aliases, I guess that would be a subject for another PR.

Please, review! :checkered_flag:

README.rst
```diff
@@ -87,7 +87,7 @@ Inside tests it is possible to change the log level for the captured
 log messages.  This is supported by the ``caplog`` funcarg::
 
     def test_foo(caplog):
-        caplog.setLevel(logging.INFO)
+        caplog.set_level(logging.INFO)
```
The-Compiler, line 90: Those documentation changes seem wrong as the API is still the old one.

The-Compiler:
LGTM apart from the comment above. :smile:

README.rst
abusalimov, line 90: The API was changed long ago: in f5c3537d70a2d95303f71e7bea4b96878a1887b1 (and tests in 948bfcf5ba6fe9969c66e06a76d9a1f3900683ca), I've only updated the docs to be in-sync with the code.

Let's consider restoring backward compatibility as a separate issue. :cactus:

eisensheng:
Dammit, I reviewed the Pull Request on a per commit base and now the comments aren't visible from here. I am still somewhat inexperienced with github from this point of view after all. Sorry!

I hope the code review wasn't too destructive, I still like the over all code quality in terms of readability, semantics and how everything is well documented.

The-Compiler:
@eisensheng your comments are actually visible here, just not with context - but that's available when clicking on the commit id.

I agree with the `_make_runtest_for` thing - I found it quite tricky to understand the first time I saw it.

abusalimov:
@eisensheng @The-Compiler Thank you for the comments! And oops, I've just noticed, that I removed the usage of `self.formatter`, which means that `--log-format` doesn't work now.

I'll try to finish fixing these till this weekend.

abusalimov:
> Dammit, I reviewed the Pull Request on a per commit base

@eisensheng That's fine, I can see your comments.

The main drawback of doing that is that rebasing will completely hide these comments. That is, IMO per-commit comments are about certain changes, for instance when a commit is too big, or you don't like a commit log message - the changes, which should gone after rebase.

abusalimov:
Hi guys,

I've fixed most of the review comments (hackish hooks and PEP8 blank lines). However, I have a problem with running tests on Python 3.5. Please see builds [19](https://travis-ci.org/eisensheng/pytest-catchlog/builds/84768895) (904b138d72160a4bb0943a5802f99218d5525400) and [20](https://travis-ci.org/eisensheng/pytest-catchlog/builds/84768985) (83eb180f1980b4dce7aabc5a5d3f5aee587d150a).

The symptoms are quite similar to these described in pytest-dev/pytest#1008, because Travis uses pytest-2.8.0 for Python 3.5 and pytest-2.6.4 for the rest. The problem is reproduced irrespective to using tox or invoking py.test directly.

@The-Compiler could you help me, please? Honestly, I'm a bit stuck.

abusalimov:
@The-Compiler I've just discovered, that replacing `catchlog` with `pytest_catchlog` in `pytest_plugins` fixes the issue for pytest-2.8:

``` diff
diff --git a/test_pytest_catchlog.py b/test_pytest_catchlog.py
index d5b7cfd..d406c45 100644
--- a/test_pytest_catchlog.py
+++ b/test_pytest_catchlog.py
@@ -1,6 +1,6 @@
 import py

-pytest_plugins = 'pytester', 'catchlog'
+pytest_plugins = 'pytester', 'pytest_catchlog'


 def test_nothing_logged(testdir):
@@ -8,7 +8,7 @@ def test_nothing_logged(testdir):
         import sys
         import logging

-        pytest_plugins = 'catchlog'
+        pytest_plugins = 'pytest_catchlog'

         def test_foo():
             sys.stdout.write('text going to stdout')
```

But this makes older version to fail with the following error:

```
...
.tox/py27/local/lib/python2.7/site-packages/_pytest/core.py:185: in register
    name, plugin, self._name2plugin))
E   ValueError: Plugin already registered: pytest_catchlog=<module 'pytest_catchlog' from '/home/eldar/work/pytest-catchlog/.tox/py27/local/lib/python2.7/site-packages/pytest_catchlog.pyc'>
E   {'terminalreporter': <_pytest.terminal.TerminalReporter instance at 0x7f6e355e4b48>, ... }
```

Shouldn't this be considered as a pytest-core regression?

The-Compiler:
Interesting... it looks to me like something about passing plugins to subprocesses run via `pytester` didn't work at all before 2.8, i.e. that `pytest_plugins` was ignored. Maybe @nicoddemus can take a look? I'm guessing something changed with https://github.com/pytest-dev/pytest/pull/860

The thing is: `pytester` already _does_ load the plugin for you, that's why it's now getting loaded twice if you fix the name. So I think all you need to do to make this work with all versions is to use `pytest_plugins = 'pytester'` once in `test_pytest_catchlog.py` and remove all other occurences of `pytest_plugins = ...`.

nicoddemus:
Thanks for the pointer @The-Compiler, I will take a look when I have some spare time.

> So I think all you need to do to make this work with all versions is to use pytest_plugins = 'pytester' once in test_pytest_catchlog.py and remove all other occurences of pytest_plugins = ....

I agree, you shouldn't need to add `catchlog` to `pytest_plugins`. When you execute the tests under tox, the plugin will be installed in the virtualenv so you don't have to explicitly add it.

abusalimov:
@The-Compiler @nicoddemus thank you, guys! Everything seems to be fine now.
@eisensheng I've mostly finished. There is an arguable change introduced in 83eb180f1980b4dce7aabc5a5d3f5aee587d150a ("travis: Don't use tox, invoke py.test directly") though that one might not like. WDYT?

The-Compiler:
>  There is an arguable change introduced in 83eb180 ("travis: Don't use tox, invoke py.test directly") though that one might not like. WDYT?

What's the benefit of that?

eisensheng:
> @eisensheng I've mostly finished. There is an arguable change introduced in 83eb180 ("travis: Don't use tox, invoke py.test directly") though that one might not like. WDYT?

A good reason to use tox is that it also allows to test the installation process of the given package. While this plugin is not a huge process and there aren't many things that can go wrong I still would say that we can benefit from this feature set. But since this project is that small I guess it can be overlooked since there isn't much that can go wrong during installation so it comes close to YAGNI. So I'm kind of so-so on this topic since both arguments are balanced in my head. Either way is fine with me.

abusalimov:
@The-Compiler Well, mostly a matter of personal preferences.

tox is good for testing against multiple Python versions at once during the development, but Travis already provides properly initialized isolated venvs for that. Running tox inside Travis creates a new (unnecessary) venv plus clutters the log output. I know that having a single point for defining test command is a good thing, but, come on, this is just a single `py.test` line (compare with e.g. [tox.ini of pytest](https://github.com/pytest-dev/pytest/blob/master/tox.ini)).

> A good reason to use tox is that it also allows to test the installation process of the given package.

@eisensheng IMO this is a good reason to _not_ use tox in favor of the explicit `pip install`. And explicit is better than implicit, right? :wink:

pytest_catchlog.py
```diff
@@ -139,10 +180,13 @@ def emit(self, record):
 class CatchLogFuncArg(object):
```
The-Compiler: "FuncArg" is basically the old concept of fixtures - see [reasoning for fixture/funcarg evolution](http://pytest.org/latest/funcarg_compare.html) - so I don't think this should be called funcarg anymore now that we use fixtures instead of `pytest_funcarg__capturelog` (also see the docstring of `__init__`).

The-Compiler:
Looks good to me apart from the nitpick I commented on. You convinced me about the tox thing, it's probably easier for such a simple project (in terms of installing it and running tests) indeed.

pytest_catchlog.py
abusalimov: @The-Compiler Thanks. I thought that "funcarg" and "fixtures" are roughly synonymous, the one just being a modern alternative to the other.

How do I call this class? `CatchLogFixture` / `CatchLog` / `Caplog` / `CaplogObject`?

/cc @eisensheng

The-Compiler: I'd just call it `CatchLog` or maybe `LogCatcher`. Though pytest calls the object behind the `capsys`/`capfd` fixtures `CaptureFixture`, so `CatchLogFixture` or `LogCaptureFixture` would certainly work as well.

abusalimov: I decided to pick `LogCaptureFixture` since the word _catch_ sounds a bit weird for me.  :airplane: :arrow_left: :arrow_right:

The-Compiler:
@eisensheng Is something still holding this back, or did you just not get to merging this yet? Would be nice to (finally) have this in and move on :wink:

eisensheng:
Sorry for the late response, I stay until late at the office during the week so I only have time during the weekend for private projects. I'll look over it again after I was able to catch some sleep and merge it right away. :wink: As far as I can remember everything should be fine now though.

PR #11, opened 2015-10-20
Make caplog.text, .records and .record_tuples properties and restore backward compatibility

abusalimov:
Slightly improve the API by making `caplog.text`, `caplog.records` and `caplog.record_tuples` properties while preserving backward compatibility, and also restore `caplog.atLevel()` and `caplog.setLevel()` back (as aliases for their snake_case counterparts).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -119,9 +119,9 @@ the contents of a message::
 
     def test_baz(caplog):
         func_under_test()
-        for record in caplog.records():
+        for record in caplog.records:
             assert record.levelname != 'CRITICAL'
-        assert 'wally' not in caplog.text()
+        assert 'wally' not in caplog.text
 
 For all the available attributes of the log records see the
 ``logging.LogRecord`` class.
@@ -133,7 +133,7 @@ given severity and message::
     def test_foo(caplog):
         logging.getLogger().info('boo %s', 'arg')
 
-        assert caplog.record_tuples() == [
+        assert caplog.record_tuples == [
             ('root', logging.INFO, 'boo arg'),
         ]
 
--- a/pytest_catchlog.py
+++ b/pytest_catchlog.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import, division, print_function
 
+import functools
 import logging
 from contextlib import closing, contextmanager
 
@@ -188,16 +189,17 @@ class LogCaptureFixture(object):
         """Creates a new funcarg."""
         self._item = item
 
+    @property
     def text(self):
         """Returns the log text."""
-
         return self.handler.stream.getvalue()
 
+    @property
     def records(self):
         """Returns the list of log records."""
-
         return self.handler.records
 
+    @property
     def record_tuples(self):
         """Returns a list of a striped down version of log records intended
         for use in assertion comparison.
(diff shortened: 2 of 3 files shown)


abusalimov:
@The-Compiler answering your comment from [here](https://github.com/eisensheng/pytest-catchlog/pull/7#issuecomment-144712274):

> Note subclassing `list` isn't a good idea as subclassing a builtin comes with many fun surprises when C code uses `list` instead of the subclass internally. Something like this would be better I think (untested):
> ...

I decided to inherit the builtin types directly nonetheless, because we actually _do_ want the object to be treated as `list`/`str` instead of the subclass. After all, we only add the `__call__` slot that is not used internally anyway.

eisensheng:
I thought about this for a while/too long and I would suggest a different course of action regarding the fix for `atLevel` and `setLevel`.

I still would like to introduce them in the 1.2 release and market it this as a bugfix to retain backwards compatibility. But I suggest to not just have this API for the sake of backward compatibility but rather to give them a proper deprecation phase and suggest to use the equivalent snake_case named method. PEP8 discourages lowered camelCase for method names and I would love to adhere to PEP8 where possible and advisable.

Please let me know what you think about this since this is blocking 1.2 a little bit. I don't want to release 1.2 and introduce the compatibility fix in another release before 2.0.

abusalimov:
@eisensheng Well, I left a TODO about a possible warning there since I didn't find the proper way for a pytest plugin to emit an API warning. Pytest itself uses some _internal functions to make its own warnings. The standard Python `warnings.warn()` doesn't seem to show anything on the screen for the `DeprecationWarning`.

The README advertises to use the proper snake_case methods anyway; and for anyone who would inspect the source code it should be obvious that mixedCase methods under the `CompatLogCaptureFixture` class calling the snake_case ones are only for compatibility. So the changes proposed by this PR don't add any confusion IMO.

To be clear, I'm not against what you suggest, I'm for emitting warnings. I just don't know how to do it the right way.

abusalimov:
@hpk42 @nicoddemus @The-Compiler Hi guys, is there an official way for a plugin to emit an API deprecation warning? We need to ask our users to stop using `mixedCase` methods in favor of the `snake_case` ones. Thanks in advance.

eisensheng:
http://pytest.org/latest/writing_plugins.html?highlight=plugins#_pytest.config.Config.warn appears to be suitable for warning the user. 

Edit:
The underlying functionality is also used to warn the user if a plugin still uses `__multicall__`.

abusalimov:
OH, SHHHame on me, how could I have missed that!? OK, will add the warnings soon.

abusalimov:
@eisensheng ready, please review.

The-Compiler:
Looks good to me - quite some magic in there, but I can't think of a cleaner way.

abusalimov:
@eisensheng ping

eisensheng:
@abusalimov sorry for the late response, I had my hands full the last weekend so I wasn't able to look after it.

Would you mind if I rip the backward compatible fixes for `atLevel` and `setLevel` out of this PR and backport them for 1.2? I would love to have this in the 1.2 release as a bug fix for the atrocity i committed and merge this PR for 2.0 since it breaks the new API introduced by me back then. (Which hopefully nobody uses.)

The-Compiler:
I don't follow - which API does this PR break?

abusalimov:
I don't either. This PR doesn't break newer `snake_case` APIs, neither it promotes the old `mixedCase` to be used, just the opposite.

@eisensheng Do you mean, you'd rather throw `at_level` away in favor of `atLevel`?

eisensheng:
Sorry, my fault, I misinterpreted the code.

abusalimov:
Thanks!

PR #15, opened 2015-11-10
Restore Python 2.6 compatibility

mnencia:
Release 1.2.0 broke Python 2.6 compatibility. As there is only one statement in the module that is not compatible, I think it is worth to alter it restoring the Python 2.6 compatibility and allowing the user to run tests on that python version for some more time.

### DIFF
--- a/pytest_catchlog.py
+++ b/pytest_catchlog.py
@@ -64,11 +64,11 @@ def catching_logs(handler, filter=None, formatter=None,
         handler.setFormatter(formatter)
     handler.setLevel(level)
 
-    with closing(handler), \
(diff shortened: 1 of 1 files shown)


eisensheng:
Thanks for pointing that out! I guess we should explicitly test against Python 2.6 on Travis CI and advertise Python 2.6 support in the `setup.py` file since 2.6 is also fully supported by Pytest.

I'll merge this one here and provide a merge request myself for the other stuff on the weekend unless someone else beats me to it. I got a little bit of work piled up anyways.

mnencia:
Created #17 to test 2.6 compatibility.

PR #20, opened 2015-11-15
Move pytest-logging functionality to pytest-catchlog.

### LINKED ISSUE #14: Integrate with --capture and -v options of pytest
There's a [pytest-logging](https://github.com/saltstack/pytest-logging) plugin by @s0undt3ch that provides a nice way to control logging through the default `-v` and `-s` (alias for `--capture=no`) options of pytest-core. The verbosity flag naturally configures the logging level, and the capturing option - whether to output live logs for all tests, or only capture them to show in case of a failure afterwards.

For instance, one can run `py.test -vv -s` to see live logs at the `INFO` level or above emitted by all tests.

It would be nice to have something like that in pytest-catchlog.

s0undt3ch:
Fixes #14


abusalimov:
Thank you!

I think your logic could and should be more integrated into the existing code. For example, I believe, it could be achieved without attaching one more handler and formatter. How about reusing `LogCaptureHandler()`, which is created in `_runtest_for()`?

s0undt3ch:
LogCaptureHandler does not write to `sys.stderr` if I'm reading it correctly...

pytest_catchlog.py
```diff
@@ -11,6 +12,29 @@
...
+# to be "passed" to the handlers
+logging.root.setLevel(1)
+
+
+class Hooks(object):  # pylint: disable=too-few-public-methods
+    """
+    Class to add new hooks to pytest
+    """
+
+    def pytest_register_logging_levels(self):
```
abusalimov: Do we really need to know all user-defined levels. What's about just calculating the desired level from the verbosity flag, using a fixed formula, like `level = logging.ERROR - 10*v`?

abusalimov:
> LogCaptureHandler does not write to `sys.stderr` if I'm reading it correctly...

Then we should teach it to, depending on the `-s`/`--capture` flag, I guess.

pytest_catchlog.py
```diff
@@ -11,6 +12,29 @@
 
 __version__ = '1.2.0'
 
+TERMINAL = py.io.TerminalWriter(sys.stderr)  # pylint: disable=no-member
+CONSOLEHANDLER = logging.StreamHandler(TERMINAL)
+# Add the handler to logging
+logging.root.addHandler(CONSOLEHANDLER)
+# The root logging should have the lowest logging level to allow all messages
+# to be "passed" to the handlers
+logging.root.setLevel(1)
```
abusalimov: This shouldn't be done at the global script level, since it effectively forbids the module to be imported by anyone, including various tools.

pytest_catchlog.py
s0undt3ch: I don't think you can get a fixed formula to account for user defined levels... However, it just made me realize that python logging module does know about these levels, so I'll see if there's a way to avoid the custom hook.

s0undt3ch:
Hmm... I'll move it...
Is there a way to tell pytest to configure this plugin before others? Even on a best effort?
We want logging setup as early as possible...

pytest_catchlog.py
abusalimov: > I don't think you can get a fixed formula to account for user defined levels...

Why not? User level are anyway somewhere between NOTSET and CRITICAL, more likely between NOTSET and DEBUG. So we could just take it into consideration, i.e. scale the range properly, for example:

```
ERROR   (40) default
WARNING (30) -v 
INFO    (20) -vv
DEBUG   (10) -vvv
CUSTOM  (5)  -vvvv
CUSTOM  (2)  -vvvvv
```

abusalimov:
> Is there a way to tell pytest to configure this plugin before others? Even on a best effort?

You could use `@pytest.mark.tryfirst` for that.

pytest_catchlog.py
abusalimov: Or just let the user to customize it through `pytest.ini` maybe...

s0undt3ch: I'd like to avoid guessing or getting close. The logging module should give use the available logging levels, though, libraries which are late imported could add log levels after the plugin has had a chance to translate verbosity to log level...

s0undt3ch:
> > LogCaptureHandler does not write to sys.stderr if I'm reading it correctly...
> 
> Then we should teach it to, depending on the -s/--capture flag, I guess.

Here's my issue with `LogCaptureHandler`, if I read the code correctly, a new handler is instantiated before each test is executed, my purpose with pytest-logging was to even provide logging for any other py.test plugins which needed it, not just while running a test.

abusalimov:
> if I read the code correctly, a new handler is instantiated before each test is executed, my purpose with pytest-logging was to even provide logging for any other py.test plugins which needed it, not just while running a test

You're right, but I'm not sure that that behavior is desired for most users. Honestly, I can't imagine a case when I would be interested in what is happening outside my tests.

Could you provide an example, please?

@The-Compiler @eisensheng, what do you think?

s0undt3ch:
As I had mentioned on the [issue filed against my plugin](https://github.com/saltstack/pytest-logging/issues/1#issuecomment-155060139), I am sometimes, actually quite often, interested on some boiler plate logging that happens when preparing the test suite, that's why I needed logging support in pytest in the first place...

s0undt3ch:
I suggest keeping 2 handlers, which already exists, which takes care of catching the log messages, and one responsible for printing them to the console. In this case, CatchLogHandler wouldn't need to subclass stream handler since we would already have a stream handler installed...

s0undt3ch:
I'm also interested on logging that happens on libraries that I use with my own library, hence me setting the root logger level to 1, so we'd get all logging from all libraries that are used by the code being tested.

abusalimov:
> I am sometimes, actually quite often, interested on some boiler plate logging that happens when preparing the test suite

catchlog handles this as well by setting up a handler that captures logs during setup phase, i.e. preparing and running necessary fixtures needed to run a test case. Isn't this what you need?

s0undt3ch:
I'm trying to see if I can "fit" your concerns into some usable code....

abusalimov:
I mean, if you just need to log some boilerplate code that fits the standard pytest fixtures concept or setup hook wrappers, then everything should work fine already.

``` python
@pytest.fixture
def fixx():
    logger.error('FIXX')

def test_foo(caplog, fixx):
    logger.info('test')
```

In this example, both log lines are captured, yet by handlers created separately.

s0undt3ch:
Yes, I read that correctly, but I'm hacking another pytest plugin which needs logging, and I'm trying to confirm if that approach will serve that purpose as well, pytest-logging was going to facilitate logging to other plugins...

s0undt3ch:
So far, hook usage was dropped...

abusalimov:
> I suggest keeping 2 handlers, which already exists, which takes care of catching the log messages, and one responsible for printing them to the console. In this case, CatchLogHandler wouldn't need to subclass stream handler since we would already have a stream handler installed...

On the second thought, it could be a good solution to keep a single handler instance attached to the logging subsystem (either streaming to `stderr` or capturing records into a buffer, depending on the cmdline options).

This reflects how the standard stdout/stderr capturing works, to some extent. Also this could reduce the overhead of configuring the logging subsystem each time a test function runs.

s0undt3ch:
Well, I actually have to disagree with you... I'll explain.

When I coded pytest-logging I was looking for an easy way to get _all_ logging, before, during and after running the test suite, for all libraries involved. This is something that is most likely to be used while developing tests or while debugging why a test started to fail, not all the time.
However, on a CI context, I'm most likely interested on the logging that happened while the test was being prep'ed till the end of it, that's why `CatchLogHandler` uses a TextIO stream(to get back the stream contents and include them in the report).
The console stream is "attached" to `sys.stderr`, you can't get back it's contents.

So either e attach 2 stream to CatchLogHandler(in case verbosity is higher than  1) and write to both streams, or, we have separate handlers(which actually serve 2 different purposes....)

abusalimov:
You seem to miss the point, I was talking on. I can understand both use-cases you described, and that's exactly what I meant by saying 

> either streaming to stderr or capturing records into a buffer, depending on the cmdline options

What I'm also trying to say is that you won't need both handlers **at once**. But I was not suggesting to merge the logic providing these two distinct flavors into a single Handler class. There likely should be two classes, though only one of them is instantiated and attached as a logging handler.

s0undt3ch:
Ok, please review the code again. No hook usage(although that breaks pypy testing so I'm skipping those), and no global instantiation...
Still 2 separate stream handlers.

Let me know if this is more likely to get merged of if there's still something else that I need to address...

s0undt3ch:
Honestly, I'd prefer to support the hooks as it previously did, but i's your call now...

s0undt3ch:
Then its the _or_ that I'm missing. There might be tests which will need
the captured log, so, I believe it shouldn't be a matter of _or_. You'll
get 2 handler, one which has the ability to output to the console in
realtime(which can't be used in tests to confirm that a log message was
issued), the other which store the log messages and is capable of allowing
a test to ensure that a log message was issued....

On 16 November 2015 at 16:12, Eldar Abusalimov notifications@github.com
wrote:

> You seem to miss the point, I was talking on. I can understand both
> use-cases you described, and that's exactly what I meant by saying
> 
> either streaming to stderr or capturing records into a buffer, depending
> on the cmdline options
> 
> What I'm also trying to say is that you won't need both handlers _at once_.
> But I was not suggesting to merge the logic providing these two distinct
> flavors into a single Handler class. There likely should be two classes,
> though only one of them is instantiated and attached as a logging handler.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/eisensheng/pytest-catchlog/pull/20#issuecomment-157084005
> .

## 

Pedro Algarvio

abusalimov:
> (which can't be used in tests to confirm that a log message was
> issued), the other which store the log messages and is capable of allowing
> a test to ensure that a log message was issued....

Yeah, its true. I meant only the part responsible for reporting logs in case of a failure, not `caplog` fixture stuff.

I have [pushed](https://github.com/eisensheng/pytest-catchlog/compare/eisensheng:4f1c636...abusalimov:214bfc5) some experimental changes that I made two month ago as part of #7. They include separating Handlers responsible for reporting and inspecting. I should have given you that links first, that would probably clarify a lot: 00b394d3600375573a295e76708f363476b517f4 and 12e7ce0c87f0b3d3a872c644cba61784479c47b0.

s0undt3ch:
Let me know if there's anything else that you want me to address differently.
Also, I'd like to include back the hooks since a library which adds log levels might be late imported and catchlog won't have access to those newly added log levels. Are you completely against?

pytest_catchlog.py
```diff
@@ -122,6 +127,15 @@ def __init__(self, config):
...
                 get_option_ini(config, 'log_format'),
                 get_option_ini(config, 'log_date_format'))
+        terminal = py.io.TerminalWriter(sys.stderr)  # pylint: disable=no-member
+        self.console = logging.StreamHandler(terminal)
+        self.console.setFormatter(self.formatter)
+        # Add the handler to logging
+        logging.root.addHandler(self.console)
+        # The root logging should have the lowest logging level to allow all
+        # messages to be "passed" to the handlers
+        logging.root.setLevel(1)
```
abusalimov: AFAIK, the lowest level is 0, i.e. `logging.NOTSET`.

pytest_catchlog.py
```diff
@@ -154,6 +168,48 @@ def pytest_runtest_teardown(self, item):
...
+            if level <= logging.NOTSET:
+                continue
+            if level in log_levels:
+                continue
+            log_levels.append(level)
+
+        log_levels = sorted(log_levels, reverse=True)
+
+        for idx, level in enumerate(log_levels):
+            handled_levels[idx + 2] = level
```
abusalimov: Honestly, it's pretty hard to follow this code, at least for me...

pytest_catchlog.py
```diff
@@ -154,6 +168,48 @@ def pytest_runtest_teardown(self, item):
...
+        # Prepare the handled_levels dictionary
+        log_levels = []
+        handled_levels = {}
+        try:
+            available_levels = logging._levelNames.keys()
+        except AttributeError:
+            # Python >= 3.4
+            # And also PyPy3-2.4.0(Python 3.2) apparently...
+            # https://travis-ci.org/eisensheng/pytest-catchlog/jobs/91417631
+            available_levels = logging._levelToName.keys()
```
abusalimov: Relying on the implementation details is a rather bad idea, trust me. Let's discuss other possible approaches to get the info on custom levels.

abusalimov:
@s0undt3ch Sorry for the delayed response.

> Also, I'd like to include back the hooks since a library which adds log levels might be late imported and catchlog won't have access to those newly added log levels. Are you completely against?

Honestly I don't like the hooks approach. But I have some thoughts about a similar ini-option:

``` ini
[pytest]
log_level_extra = 
    TRACE           # infer proper number of -v from the level value
    -vvvv:GARBAGE   # or map the verbosity flag explicitly
```

What do you think?

pytest_catchlog.py
s0undt3ch: I have another approach, add a hook and let the user define additional logging levels, which, I believe is the most future proof way of doing this. We add a mapping of the log levels that we handle, and, if no other logging levels are defined.....

s0undt3ch:
I like to keep my plugins customization in my `conftest.py` and, only as a last resort, use `pytest.ini`(I actually only use pytest.ini for data which should not be made public, like credentials).... `log_level_extra` is adding extra functionality(extra log levels), not configuring existing functionality....

s0undt3ch:
Updated the PR to address _some_ of your concerns...

abusalimov:
Sorry, but I have to disagree with you.

> (I actually only use pytest.ini for data which should not be made public, like credentials).... 

`pytest.ini` is usually under VCS control, I can't imagine putting sensitive data there.

> `log_level_extra` is adding extra functionality(extra log levels), not configuring existing functionality....

What extra functionality does it add per se? It is really tweaking an existing feature, which in turn is solely a usability thing: it just tells how to interpret `-vvv` argument, doesn't it?

Anyway you still able to customize this thing from the `conftest.py`, I guess something like that might work:

``` py
def pytest_configure(config):
    config.addinivalue_line('log_level_extra', 'TRACE')
```

I would like to hear comments on that from @eisensheng and @The-Compiler.

s0undt3ch:
I use pytest in a CI context, pytest.ini get's added "manually" to include credentials so we don't have to pass them using environmental variables or using arguments to pytest(which are exposed in the CI context)

eisensheng:
> I would like to hear comments on that from @eisensheng and @The-Compiler.

`_levelNames` is not documented anywhere on https://docs.python.org/ so in my opinion it's really a bad idea to relay on this. Thus specifying level names in the `pytest.ini` is a good idea in my opinion. Maybe we can also abuse the `setup.cfg` file to let users specify them there. This would also cover @s0undt3ch's use case.

s0undt3ch:
What's the story against hooks? It's python, there's no parsing involved....

abusalimov:
Mainly because that's too powerful and generic solution for such a simple and optional usability feature.

Just compare:

``` py
class CatchlogAddon(object):  # we want it to degrade gracefully
                              # if the plugin is not installed, hence
                              # this optional plugin to pass hook validation

    def pytest_register_logging_levels(self):
        if hasattr(logging, 'TRACE'):
            return logging.TRACE

def pytest_configure(config):
    if config.pluginmanager.hasplugin('pytest_catchlog'):
        config.pluginmanager.register(CatchlogAddon())
```

vs.

``` ini
[pytest]
log_level_extra = TRACE
```

OR if you don't like ini files, here's a special option for your use case:

``` py
def pytest_configure(config):
    config.addinivalue_line('log_level_extra', 'TRACE')
```

Which one shows the intention more clearly? Which one would you use more likely? Is there anything that you can't achieve through ini file, and need a hook for that?

All pytest hooks are used to override or extend certain aspects of a test session lifecycle. And for instance, there's no hook named `pytest_register_markers`, instead one should use ini file to declare necessary markers. And if for some reason you need to do that dynamically, pytest [advises](https://pytest.org/latest/example/markers.html#custom-marker-and-command-line-option-to-control-test-runs) to use `config.addinivalue_line()`.

s0undt3ch:
fine, fine... :)

eisensheng:
I agree with @abusalimov here. @s0undt3ch, you also force every other user to write this amount of boilerplate code just so to define new logging levels. It would also be difficult to explain to the user why the inconsistency compared to configuring other aspects of the plugin is necessary for this feature.

s0undt3ch:
I'll update this PR to support ini

eisensheng:
Thanks. :+1:

s0undt3ch:
Done.

s0undt3ch:
Is there anything still missing?

pytest_catchlog.py
```diff
@@ -71,10 +72,11 @@ def catching_logs(handler, filter=None, formatter=None,
                 yield handler
 
 
-def add_option_ini(parser, option, dest, default=None, help=None):
-    parser.addini(dest, default=default,
+def add_option_ini(parser, option, dest, type=None, action=None, default=None, help=None):
+    parser.addini(dest, default=default, type=type,
```
abusalimov, line 79: What's that for?

pytest_catchlog.py
```diff
@@ -111,6 +120,8 @@ def pytest_configure(config):
 class CatchLogPlugin(object):
     """Attaches to the logging module and captures log messages for each test.
     """
+    DEFAULT_LOG_FORMAT = '%(filename)-25s %(lineno)4d %(levelname)-8s %(message)s'
+    DEFAULT_DATE_FORMAT = '%H:%M:%S'
```
abusalimov: May be extract these to the module level?

pytest_catchlog.py
```diff
@@ -118,10 +129,20 @@ def __init__(self, config):
         The formatter can be safely shared across all handlers so
         create a single one for the entire test session here.
         """
+        self.config = config
```
abusalimov: Do you really need it? Better pass it as argument to `self.configure_console_handler`.

pytest_catchlog.py
```diff
@@ -154,6 +175,73 @@ def pytest_runtest_teardown(self, item):
...
+            logging.WARNING,
+            logging.INFO,
+            logging.DEBUG,
+            logging.NOTSET
+        ])
+        for level in get_option_ini(self.config, 'log_extra_levels'):
+            try:
+                level_num = int(level)
+            except ValueError:
+                level_num = logging.getLevelName(level)
```
abusalimov: Does this do what is intended? Config value is a string like `"TRACE"`, whereas `getLevelName()` expects a level _number_, not it's name.

There should be something like `int(getattr(logging, get_option_ini(...), None))`, I suppose.

pytest_catchlog.py
s0undt3ch, line 79: To tell config that the ini setting is args and not a string(which would require splitting and parsing).

pytest_catchlog.py
s0undt3ch: Sure.

pytest_catchlog.py
s0undt3ch: Ok.

pytest_catchlog.py
abusalimov, line 79: Ah, I see. I misinterpreted this as a default value for the argument first, dumb highlighting.
Thanks for clarifying that!

pytest_catchlog.py
s0undt3ch: If the value is a string, it must be a known logging level and getLevelName returns the level num for that level name.

s0undt3ch: Don't know if that's publicly documented but that behavior exists in py2 and py3....

abusalimov: That's true only for old Python versions. Here're [docs](https://docs.python.org/3.4/library/logging.html#logging.getLevelName) for Python 3.4:

> _Changed in version 3.4:_ In Python versions earlier than 3.4, this function could also be passed a text level, and would return the corresponding numeric value of the level. This undocumented behaviour was considered a mistake, and was removed in Python 3.4, but reinstated in 3.4.2 due to retain backward compatibility.

s0undt3ch: Oh, and the config value(s)(one per line) can be either strings(level name) or int's(level nums).

abusalimov: I guess, this should do the trick then:

``` py
for level in get_option_ini(self.config, 'log_extra_levels'):
    try:
        level_num = int(getattr(logging, level, level))
    except ValueError:
        raise pytest.UsageError("{0!r} is not a valid level ...".format(level))
```

s0undt3ch:
I think I've addressed all your concerns @abusalimov

s0undt3ch:
The extra commits which are not mine is because I pulled changes from this repo's master in order to rebase and meld some commits...

s0undt3ch:
Ignore my previous comment. I had pulled from master when I should have pulled from develop.... Fixed.

s0undt3ch:
Anything else that needs to be addressed?

s0undt3ch:
Just _dropping a ping_ to get some status feedback on this PR.

abusalimov:
Yeah, things don't move fast enough, sorry for that. I still have no time to review your tests. Anyway, we have to wait until #21, #22 and #23 are merged as part of a patch release.

pytest_catchlog.py
```diff
@@ -122,6 +134,15 @@ def __init__(self, config):
         self.formatter = logging.Formatter(
                 get_option_ini(config, 'log_format'),
                 get_option_ini(config, 'log_date_format'))
+        terminal = py.io.TerminalWriter(sys.stderr)  # pylint: disable=no-member
```
abusalimov: BTW, what is the rationale of wrapping `stderr` with `TerminalWriter`? AFAIK, `logging.StreamHandler` works well with `stderr` directly.

pytest_catchlog.py
```diff
@@ -122,6 +134,15 @@ def __init__(self, config):
...
+        self.console = logging.StreamHandler(terminal)
```
abusalimov: What do you think about renaming `console` to just `handler`? I would appreciate that, because #25 [makes](https://github.com/abusalimov/pytest-catchlog/commit/4121487982a1b39937f9daa04cfc94c016e201bc#diff-bf8049f1ff38ea40a0da2b88b46ee051R129) use of a single shared handler for both capturing and live streaming. And since we decided to merge your PR first, I'd like to keep the further diff as concise as possible.

pytest_catchlog.py
```diff
@@ -122,6 +134,15 @@ def __init__(self, config):
...
+        self.console.setFormatter(self.formatter)
+        # Add the handler to logging
+        logging.root.addHandler(self.console)
```
abusalimov: If someone attaches a handler, there must be a point where it is eventually detached. The logging subsystem can't track loading/unloading of plugins by itself unfortunately, but it is necessary to provide reentrancy (at least to be able to run `pytester` in-process).

I would suggest the same approach as used in abusalimov/pytest-catchlog@4121487982a1b39937f9daa04cfc94c016e201bc:

``` diff
@@ -123,13 +123,27 @@ def __init__(self, config):
+        handler = logging.StreamHandler()
+        handler.setFormatter(self.formatter)
+
+        self.handler = handler
+
+    @pytest.mark.hookwrapper
+    def pytest_runtestloop(self, session):
+        """Runs all collected test items."""
+        with catching_logs(self.handler):
+            yield  # run all the tests
+
```

pytest_catchlog.py
```diff
@@ -154,6 +175,72 @@ def pytest_runtest_teardown(self, item):
...
+            logging.DEBUG,
+            logging.NOTSET
+        ])
+        for level in get_option_ini(config, 'log_extra_levels'):
+            try:
+                level_num = int(getattr(logging, level, level))
+            except ValueError:
+                # Python logging does not recognise this as a logging level
+                raise pytest.UsageError(
+                    '\'{0}\' is not recognized as a logging level name. Please '
```
abusalimov: PEP8 suggests to use "double quotes" to avoid `\'` escaping in such cases.

pytest_catchlog.py
```diff
@@ -154,6 +175,72 @@ def pytest_runtest_teardown(self, item):
...
+                    'consider passing the logging level num instead.'.format(
+                        level
+                    )
+                )
+            available_levels.add(level_num)
+
+        for level in available_levels:
+            if level > logging.WARN:
```
abusalimov: Should be `logging.WARNING` as above probably...

pytest_catchlog.py
```diff
@@ -154,6 +175,72 @@ def pytest_runtest_teardown(self, item):
...
+                # Log levels lower than NOTSET, we're not interested
+                continue
+            if level in log_levels:
+                # We already know about this log level
+                continue
+            log_levels.append(level)
+
+        # Reverse the list because we're interested on higher logging levels
+        # first
+        log_levels = sorted(log_levels, reverse=True)
```
abusalimov: It is a matter of preferences, but this could also be handled by the following one-liner:

``` py
log_levels = sorted(set(level for level in available_levels
                        if logging.NOTSET <= level <= logging.WARNING),
                    reverse=True)
```

pytest_catchlog.py
```diff
@@ -154,6 +175,72 @@ def pytest_runtest_teardown(self, item):
...
+        max_verbosity = len(handled_levels) + 1
+        if verbosity > 1:
+            if verbosity in handled_levels:
+                log_level = handled_levels[verbosity]
+            elif verbosity >= max_verbosity:
+                log_level = handled_levels[max_verbosity]
+            else:
+                log_level = handled_levels[min_verbosity]
+            self.console.setLevel(log_level)
+        else:
```
abusalimov: i.e. verbosity == 1 or less (then set to `logging.FATAL`). But the comment above claims that `-v` corresponds to the `WARNING` level. Which is right?

pytest_catchlog.py
```diff
@@ -154,6 +175,72 @@ def pytest_runtest_teardown(self, item):
...
+        # Build a dictionary mapping of verbosity level to logging level
+        # -v    WARN
+        # -vv   INFO
+        # -vvv  DEBUG
+        # - ... etc
+        for idx, level in enumerate(log_levels):
+            handled_levels[idx + 2] = level
+
+        # Set the console verbosity level
+        min_verbosity = 2  # -v  - WARN loggin level
```
abusalimov: What's the matter with this magic `2`? This is rather confusing, to be honest...
Can't we stick to zero-based indices somehow?

pytest_catchlog.py
```diff
@@ -154,6 +175,72 @@ def pytest_runtest_teardown(self, item):
...
+        # first
+        log_levels = sorted(log_levels, reverse=True)
+
+        # Build a dictionary mapping of verbosity level to logging level
+        # -v    WARN
+        # -vv   INFO
+        # -vvv  DEBUG
+        # - ... etc
+        for idx, level in enumerate(log_levels):
+            handled_levels[idx + 2] = level
```
abusalimov: This can be made simpler:

``` py
handled_levels = dict(enumerate(log_levels, start=2))
```

(ideally, without this start=2, i.e. zero-based)

pytest_catchlog.py
```diff
@@ -154,6 +175,72 @@ def pytest_runtest_teardown(self, item):
...
+            try:
+                level_num = int(getattr(logging, level, level))
+            except ValueError:
+                # Python logging does not recognise this as a logging level
+                raise pytest.UsageError(
+                    '\'{0}\' is not recognized as a logging level name. Please '
+                    'consider passing the logging level num instead.'.format(
+                        level
+                    )
+                )
```
abusalimov: I'm not sure, but these parens also don't look pythonic. Please correct me if I'm wrong.

abusalimov:
If I read correctly, the last part of `configure_console_handler()` performs the following:

``` py
def configure_console_handler(config):
    # prepare available_levels, but do not exclude logging.CRITICAL

    log_levels = set(level for level in available_levels
                     if logging.NOTSET <= level <= logging.CRITICAL))

    handled_levels = dict(enumerate(sorted(log_levels, reverse=True)))

    verbosity = clamp(config.getoption('-v'), 0, len(handled_levels) - 1)
    log_level = handled_levels[verbosity]
    self.console.setLevel(log_level)
```

Where `clamp()` is defined as:

``` py
def clamp(n, minn, maxn):
    if n < minn:
        return minn
    elif n > maxn:
        return maxn
    else:
        return n
```

My suggestion is to extract this logic into `pytest_configure`, assign the resulting `log_level` as an attribute of `config` and then reuse it inside the constructor of `CatchLogPlugin`.

What do you think?

pytest_catchlog.py
s0undt3ch: I wrote that part a while back, In don't remember why now. I can switch that to sys.stderr if you prefer....

pytest_catchlog.py
s0undt3ch: Sure, I'll rename.

abusalimov: Thank you!

pytest_catchlog.py
s0undt3ch: I'll try that approach and check if I achieve the same results...

pytest_catchlog.py
s0undt3ch: I'll update.

s0undt3ch: I'm only a PEP8 follower upnto a point, but I'll also adjust this...

s0undt3ch: The above was to reply to the observation bellow...

pytest_catchlog.py
s0undt3ch: Yeah, and I even think WARN is going to  be deprecated..... At least the method is....

pytest_catchlog.py
s0undt3ch: I agree, its a matter of preference and respective readability...

pytest_catchlog.py
s0undt3ch: Interesting.... Thanks!

pytest_catchlog.py
s0undt3ch: The 2 is because the logging kicks in when verbose is 2, so, I either subtract 2 every time I lookup the level, or it starts at 2 and no subtraction happens....

pytest_catchlog.py
s0undt3ch: I'll need to update the comment

s0undt3ch:
K, I'll also do that.

Once I'm done updating, I'll post a comment.
Thanks for the review and the learning experience, yes, today is a better day because I learned something new.....

abusalimov:
You're very welcome, thank you for the efforts!

s0undt3ch:
I've updated the PR with your suggestions. I did however left some of the logic in a more verbose and commented way to better enlighten someone who's reading that for the first time. If you prefer, I'll shorten that too.

pytest_catchlog.py
```diff
@@ -92,19 +97,93 @@ def pytest_addoption(parser):
...
+                # We already know about this log level
+                continue
+            log_levels.append(level)
+
+        # Build a dictionary mapping of verbosity level to logging level
+        # verbosity=0         FATAL    (no pytest output is shown and FATAL
+        #                               log messages are displayd)
+        # verbosity=1   -v    CRITICAL (pytest verbosity kicks in, but only
+        #                               log messages with higher or equal
+        #                               level to CRITICAL are shown)
```
abusalimov: Wait, but `FATAL` is just an alias for `CRITICAL`...

pytest_catchlog.py
```diff
@@ -92,19 +97,93 @@ def pytest_addoption(parser):
...
+        min_verbosity = min(handled_levels)
+        max_verbosity = max(handled_levels)
+        if verbosity in handled_levels:
+            cli_handler_level = handled_levels[verbosity]
+        elif verbosity >= max_verbosity:
+            cli_handler_level = handled_levels[max_verbosity]
+        else:
+            cli_handler_level = handled_levels[min_verbosity]
+
+    setattr(config, '_catch_log_cli_handler_level', cli_handler_level)
```
abusalimov: No need to use `setattr` here, just assign it directly:

``` py
config._catch_log_cli_handler_level = cli_handler_level
```

s0undt3ch: K, I'll update.

pytest_catchlog.py
```diff
@@ -122,6 +201,11 @@ def __init__(self, config):
         self.formatter = logging.Formatter(
                 get_option_ini(config, 'log_format'),
                 get_option_ini(config, 'log_date_format'))
+        self.handler = logging.StreamHandler(sys.stderr)
+        self.handler.setFormatter(self.formatter)
+        # The root logging should have the lowest logging level to allow all
+        # messages to be "passed" to the handlers
+        logging.root.setLevel(logging.NOTSET)
```
abusalimov: This is redundant, `catching_logs(...)` takes care of this anyway.

pytest_catchlog.py
s0undt3ch: Err... Then that I didn't include it in the first implementation.... Thanks. I'll fix.

pytest_catchlog.py
s0undt3ch: For the specific handler and logger, right? Don't know if we'll get 3rd party logging without it.... I'll try.

pytest_catchlog.py
abusalimov: But why is the highest level not included at the first place? It would eliminate the need to start counting from 1 in the `enumerate` call below, and let us use zero-based indices I was talking about yesterday...

pytest_catchlog.py
abusalimov: Yes, `catching_logs` adjusts the level of a given logger (root in our case) in case a particular handler requires a lesser level.

s0undt3ch:
Updated.

abusalimov:
Hi,

Could you join our gitter chat now? I have some though I'd like to discuss
with you.

Regards,
Eldar

pytest_catchlog.py
```diff
@@ -92,19 +97,91 @@ def pytest_addoption(parser):
...
+                   help='extra logging levels that verbosity should be sensible to.')
 
 
 def pytest_configure(config):
     """Always register the log catcher plugin with py.test or tests can't
     find the  fixture function.
     """
+    verbosity = config.getoption('-v')
+    if verbosity < 1:
+        cli_handler_level = logging.FATAL
```
abusalimov: I guess `logging.CRITICAL` is preferred over it's alias. The official docs used to use `CRITICAL`.

pytest_catchlog.py
```diff
@@ -92,19 +97,91 @@ def pytest_addoption(parser):
...
+    elif verbosity >= 1:
+        # Prepare the handled_levels dictionary
+        log_levels = []
+        available_levels = set([
+            #logging.FATAL,  # This will be the default console level if no -v is passed
```
abusalimov: And still...

> why is the highest level not included at the first place? It would eliminate the need to start counting from 1 in the enumerate call below, and let us use zero-based indices I was talking about yesterday.

abusalimov:
What do think about moving the level range checking logic into the first loop, which parses ini-option?

``` py
def pytest_configure(config):
    available_levels = set([
        logging.NOTSET,
        logging.DEBUG,
        logging.INFO,
        logging.WARNING,
        logging.ERROR,
        logging.CRITICAL,
    ])
    for level in get_option_ini(config, 'log_extra_levels'):
        try:
            level_num = int(getattr(logging, level, level))
        except ValueError:
            raise pytest.UsageError('...')
        else:
            if not (logging.NOTSET <= level_num <= logging.CRITICAL):
                config.warn("'{0}' is ignored as not being in the valid range")
                continue

        available_levels.add(level_num)

    log_levels = sorted(available_levels)
```

After that, adjusting the level based on the verbosity becomes extremely simple and straightforward:

``` py
    base_idx = log_levels.index(logging.WARNING)  # This is the default level.
    verbosity = config.getoption('-v')

    level_idx = base_idx - verbosity  # also need to clip it into a valid range

    cli_handler_level = log_levels[level_idx]
```

And here's how I would do the range clipping (any of these):
- Using a separate function suggested above:
  
  ``` py
  level_idx = clamp(base_idx - verbosity, 0, len(log_levels)-1)
  ```
- OR using a one-liner:
  
  ``` py
  level_idx = min(max(base_idx - verbosity, 0), len(log_levels)-1)
  ```
- OR the same thing, but in three lines:
  
  ``` py
  level_idx = base_idx - verbosity
  level_idx = max(level_idx, 0)
  level_idx = min(level_idx, len(log_levels)-1)
  ```
- OR an imperative one:
  
  ``` py
  level_idx = base_idx - verbosity
  if level_idx < 0:
      level_idx = 0
  elif level_idx > len(log_levels)-1:
      level_idx = -1
  ```
- OR more elegant:
  
  ``` py
  level_idx = base_idx - verbosity
  if not (0 <= level_idx < len(log_levels)):
      level_idx = (0 if level_idx < base_idx else -1)
  ```
  
  Personally, I like this one the most, because it clearly shows the intention just through the code itself.
- OR more extremal:
  
  ``` py
  level_idx = base_idx - verbosity
  if not (0 <= level_idx < len(log_levels)):
      level_idx = -(level_idx > base_idx)  # 0 or -1
  ```

s0undt3ch:
`CRITICAL` is now included and the logic is zero index based.

s0undt3ch:
Updated(without the clamping logic), see if you still prefer it in another way.

pytest_catchlog.py
```diff
@@ -92,19 +97,81 @@ def pytest_addoption(parser):
...
+                    level))
+        else:
+            if not (logging.NOTSET <= level_num <= logging.CRITICAL):
+                raise pytest.UsageError(
+                    "'{0}' is ignored as not being in the valid logging "
+                    "levels range: NOTSET({1}) - CRITICAL({2})".format(
+                        level,
+                        logging.NOTSET,
+                        logging.CRITICAL))
+                continue
```
abusalimov: Excess (`raise` above)

pytest_catchlog.py
```diff
@@ -92,19 +97,81 @@ def pytest_addoption(parser):
...
                    help='log format as used by the logging module.')
     add_option_ini(parser,
                    '--log-date-format',
-                   dest='log_date_format', default=None,
+                   dest='log_date_format',
+                   default=DEFAULT_DATE_FORMAT,
                    help='log date format as used by the logging module.')
+    add_option_ini(parser,
+                   '--log-extra-level',
+                   dest='log_extra_levels',
```
abusalimov: What's about renaming it to `--log-level-extra` and 'log_level_extra' respectively? These would look more consistent with `--log-level` that I plan to add later.

@The-Compiler What would you say?

test_pytest_catchlog.py
```diff
@@ -274,3 +277,320 @@ def test_foo(caplog):
...
+    # make sure that that we get a '0' exit code for the testsuite
+    assert result.ret == 0
+
+
+def test_logging_level_trace(testdir):
+    with open(os.path.join(testdir.tmpdir.strpath, 'conftest.py'), 'w') as wfh:
+        wfh.write(textwrap.dedent('''
+            def pytest_configure(config):
+                config.addinivalue_line('log_extra_levels', '5')
+        '''))
```
abusalimov: There's a `testdir.makeconftest()` method just for this case. ;-)

pytest_catchlog.py
s0undt3ch: Don't follow the excess part, its still a usage error  and your `config.warn` suggestion seemed too shy and easy to miss. You'd need to pass `-rw` to see it right?

And, we should stick with the same error reporting for both situations. So either the usage error, or the warn. Unless I'm missing something, which is also possible.

So, which one?

test_pytest_catchlog.py
s0undt3ch: Thanks. I'll refactor.

pytest_catchlog.py
s0undt3ch: Sure, I'll rename.

pytest_catchlog.py
abusalimov: No, I just meant that `continue` after `raise` is never reached anyway.

s0undt3ch: Oh. Oops. Yeah, absolutely right.

abusalimov: > And, we should stick with the same error reporting for both situations. So either the usage error, or the warn.

Makes sense to me. I agree with your choice.

s0undt3ch:
Updated.

abusalimov:
Let's summarize remaining issues:
- [ ] The default level is `logging.CRITICAL`, and basically only `-v` (i.e. non-negative verbosity) has an effect (should also handle `-q`). Possible solutions are proposed [above](https://github.com/eisensheng/pytest-catchlog/pull/20#issuecomment-164148687)
- [ ] Verbosity only affects the level of live logs. It is unclear whether we also need to adjust the capturing level
- [ ] When live logs are not enabled (the default behavior when no args are passed), logs are captured both as usual and also through the `stderr`, i.e. duplicated. #24 and #25 fix this
- [ ] _(feature)_ Setting the log level explicitly through `--log-level` (overrides `-vvv`)
- [ ] README

I think most of these (maybe except the README) can and should be addressed separately, after merging this PR.

@s0undt3ch The one last thing I would ask you for is to squash the whole changeset into a single commit, or few commits like feature+tests+readme, as you wish.

I'd also like to hear final comments from @eisensheng and @The-Compiler on this epic PR.

pytest_catchlog.py
The-Compiler, line 79: Shouldn't `action` be passed to `addini` as well?

EDIT: Err, no, it gets passed to `addoption` - nevermind.

pytest_catchlog.py
```diff
@@ -92,19 +97,80 @@ def pytest_addoption(parser):
...
+    # verbosity=0         CRITICAL (no pytest output is shown and CRITICAL
+    #                               log messages are displayd)
+    # verbosity=1   -v    ERROR    (pytest verbosity kicks in, but only
+    #                               log messages with higher or equal
+    #                               level to ERROR are shown)
+    # verbosity=2   -vv   WARNING  (start showing log messages with a
+    #                               higher or equal level to WARNING)
+    # verbosity=4   -vvv  INFO
+    # - ... etc
+    handled_levels = dict(enumerate(sorted(available_levels, reverse=True)))
```
The-Compiler, line 160: I haven't followed the previous discussion, but intuitively to me it'd make sense to show `CRITICAL` and `ERROR` without `-v` (maybe even `WARNING`), and then have `-v` or `-vv` for `INFO`. Not displaying errors without `-v` sounds wrong, IMHO.

abusalimov:
@s0undt3ch any news?

s0undt3ch:
Sorry, this Xmas season is kind of busy for me. I'll try to do what you asked today or tomorrow..

abusalimov:
Thanks! If you're really that busy, I could fix the remaining issues after merging into `develop`. I would appreciate though if you squash all commits into a single one or two.

s0undt3ch:
17 months bold twins and the whole family coming over for Xmas...yeah, busy.

Anyway, squashed. If you could take a look at the remaining issues, I'd be grateful.

Thanks! And merry XMas!

abusalimov:
Merry XMas!
Thank you. I'll try fix the rest in few days or a week.

s0undt3ch:
I found a need to log to a file all logging that is happening, at a different level that what's being shown in the console.
Should I add a commit to this pull request? Prefer it in a separate PR? part of the required functionality is already present in this PR...

abusalimov:
I think, making a separate PR is a better option anyway...

s0undt3ch:
So, would everyone be happy if:
1. we stopped relying on `-v` or `-q` and  start using `--log-level` explicitly instead?
2. Default to ERROR logging level

Issue #12, opened 2015-10-30
Document `atLevel` and `setLevel` rename


ulope:
It seems that during the fork the `atLevel` and `setLevel` methods of the `caplog` fixture have been de-camelized to `at_level` and `set_level`. This however is not mentioned in the README.

abusalimov:
Thanks for reporting this!

We actually have PR #11 addressing backward compatibility issues, it should be merged soon, I hope.

Probably one of the next steps could be to list all differences between the fork and the original plugin. FWIW, there is #4 for that.

The-Compiler:
I think this can be closed as there's #4?

eisensheng:
Maybe. I forgot to actually document that `atLevel` and `setLevel` are going through a deprecation phase. This should be noted though. I guess this will go in a Patch release before I close the issue.

eisensheng:
1.2.1 has been released with a warning regarding the `atLevel` and `setLevel` methods.

PR #21, opened 2015-11-15
compat: Let legacy clients modify caplog.records()

### LINKED ISSUE #18: #11 introduces subtle incompatibility
Before release 1.2.0 you were allowed to manipulate the list of logged events by changing the result of caplog.records().

After the introduction of the compatibility layer, such operation has no effect as the list returned by caplog.records (with or without ()) is a copy of the original one.

So, if you want to "reset" the caplog status you have to edit caplog.handler.records.

I can live with it, but I think it should be documented at least. Also I'd appreciate a caplog.reset() function.

abusalimov:
Some (maybe lots of) existing projects using pytest-capturelog, used to flush records buffer through a direct access to the list:

```
del caplog.records()[:]
```

This is the only available way to do that so far, in fact.

Fixes #18 ("#11 introduces subtle incompatibility"). Thanks @mnencia for reporting.

### DIFF
--- a/pytest_catchlog.py
+++ b/pytest_catchlog.py
@@ -246,7 +246,9 @@ class CallablePropertyMixin(object):
 
         @functools.wraps(func)
         def getter(self):
-            ret = cls(func(self))
+            naked_value = func(self)
+            ret = cls(naked_value)
+            ret._naked_value = naked_value
             ret._warn_compat = self._warn_compat
             ret._prop_name = func.__name__
             return ret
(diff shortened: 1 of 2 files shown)


abusalimov:
We probably need to add a `reset`/`flush` method to `caplog` in future, but for now it's worth just to restore broken backward compatibility.

@eisensheng I would suggest this to go into v1.2.1, WDYT?

eisensheng:
> @eisensheng I would suggest this to go into v1.2.1, WDYT?

Yeah, it would be nice to have this in 1.2.1 since one could argue this was public API and it broke in 1.2.0. I guess I'll take care of a proper sphinx based documentation for today to actually document what's public and what not.

abusalimov:
What's about addressing #12 as well? It could be fixed in the same patch release.

eisensheng:
> What's about addressing #12 as well? It could be fixed in the same patch release.

Yes, definitely. I'll provide the changes to the documentation to finally close that ticket.

PR #23, opened 2015-11-15
Deprecation warnings

eisensheng:
This adds a longer paragraph regarding the deprecation phase for the changed methods. The `CHANGES.rst` file will also be included in the `long_description` for the `setup.py` file to warn users only reading the PYPI page.

Would be nice to have this reviewed before it lands in develop. English is not my mother language after all. I hate to state this but this makes it a requirement that other review my text IMHO. Thanks!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,23 @@ Yet to be released.
 
 - [Bugfix] #15 #17 - Restore Python 2.6 compatibility. (Thanks to Marco Nenciarini!)
 
+.. attention::
+    Deprecation warning: the following objects (i.e. functions, properties)
+    are slated for removal in the next major release.
+
+    - ``caplog.at_level`` and ``caplog.set_level`` should be used instead of
+      ``caplog.atLevel`` and ``caplog.setLevel``.
+
+      The methods ``caplog.atLevel`` and ``caplog.setLevel`` are still
+      available but deprecated and not supported since they don't follow
+      the PEP8 convention for method names.
+
+    - ``caplog.text``, ``caplog.records`` and
+      ``caplog.record_tuples`` were turned into properties.
+      They still can be used as regular methods for backward compatibility,
+      but that syntax is considered deprecated and scheduled for removal in
+      the next major release.
+
 
 Version 1.2
 -----------
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,8 @@ setup(name='pytest-catchlog',
       version=_get_version(),
       description=('py.test plugin to catch log messages.'
                    ' This is a fork of pytest-capturelog.'),
-      long_description=_read_text_file('README.rst'),
+      long_description='\n'.join([_read_text_file('README.rst'),
+                                  _read_text_file('CHANGES.rst'), ]),
       author='Arthur Skowronek (Fork Author)',  # original author: Meme Dough
       author_email='eisensheng@mailbox.org',
       url='https://github.com/eisensheng/pytest-catchlog',


CHANGES.rst
```diff
@@ -12,6 +12,22 @@ Yet to be released.
 
 - [Bugfix] #15 #17 - Restore Python 2.6 compatibility. (Thanks to Marco Nenciarini!)
 
+.. attention::
+    Deprecation warnings, these symbols are slated for removal in
+    the next major release.
+
+    - ``caplog.atLevel`` for ``caplog.at_level`` and
+      ``caplog.setLevel`` for ``caplog.setLevel`` are still available
+      but not supported and deprecated since they don't follow the
```
abusalimov: I guess, _deprecated and not supported_ would be less ambiguous. :)

CHANGES.rst
```diff
@@ -12,6 +12,22 @@ Yet to be released.
...
+      PEP8 convention for method names.
+
+    - ``caplog.text``, ``caplog.records`` and
+      ``caplog.record_tuples`` were turned into properties.
```
abusalimov, line 27: M.b. `s/were/have been/` (not sure)?

CHANGES.rst
```diff
@@ -12,6 +12,22 @@ Yet to be released.
...
+      They still can be used as regular methods for backwards
+      compatibility.  They will lose the backward compatibility
```
abusalimov: Will they? I would rather use less strong statement, like _They still can be used as regular methods for backward compatibility, but that syntax is considered deprecated and scheduled for removal in the next major release._

abusalimov:
I'm not a ~~lawyer~~ native either, but it looks good to me except for few minor notes above.

CHANGES.rst
```diff
@@ -12,6 +12,22 @@ Yet to be released.
 
 - [Bugfix] #15 #17 - Restore Python 2.6 compatibility. (Thanks to Marco Nenciarini!)
 
+.. attention::
+    Deprecation warnings, these symbols are slated for removal in
+    the next major release.
```
The-Compiler: This sentence reads weird IMHO (it starting with "Deprecation warnings, ..."). I also don't think "symbols" is a common term for Python.

What about: "Deprecation warning: These functions and attributes are slated for ..."?

CHANGES.rst
```diff
@@ -12,6 +12,22 @@ Yet to be released.
...
+
+    - ``caplog.atLevel`` for ``caplog.at_level`` and
+      ``caplog.setLevel`` for ``caplog.setLevel`` are still available
```
The-Compiler: The second `setLevel` should be `set_level`

The-Compiler: This also reads kind of hard, IMHO - maybe just mention "`caplog.at_level` and `caplog.set_level` should be used instead" after "they don't follow the PEP8 convention for method names."?

CHANGES.rst
The-Compiler, line 27: I think both works,  but I'm not a nativ speaker either.

CHANGES.rst
The-Compiler: +1, sounds better to me

eisensheng:
I processed your feedback and reworked the wording a bit. I would appreciate it if someone could read over it again so I can merge this together with the other Pull Requests and Release 1.2.1 from it. Thanks! :sunglasses:

CHANGES.rst
```diff
@@ -12,6 +12,23 @@ Yet to be released.
 
 - [Bugfix] #15 #17 - Restore Python 2.6 compatibility. (Thanks to Marco Nenciarini!)
 
+.. attention::
+    Deprecation warnings, these objects (i.e. functions, properties) are
+    slated for removal in the next major release.
```
The-Compiler: I still think starting a sentence with "Deprecation warnings, these objects ..." makes no sense :wink:

I think "Deprecation warning: These objects ..." would be better.

The-Compiler:
LGTM apart from the comment above.

I think #24/#25/#26 still need some more testing though. No idea in what state the other PRs are right now.

abusalimov:
No-no-no, these must **not** go into v1.2.1.
I'll add more tests there once I finish preparing the performance changeset I mentioned in #26.

eisensheng:
Hopefully I got it right this time.

abusalimov:
I think, it's worth squashing your follow-up changes into the first commit.

PR #25, opened 2015-11-20
Live logs (-s / --capture=no)

### LINKED ISSUE #14: Integrate with --capture and -v options of pytest

abusalimov:
Partially resolves #14 ("Integrate with --capture and -v options of pytest"), in a slightly more natural and efficient way than proposed in #20 ("Move pytest-logging functionality to pytest-catchlog.")

Note: this includes changes proposed in PR #24, **DO NOT MERGE** until #24 is accepted. Also this lacks a proper description in the README.
### TODO
- [ ] README
- [ ] `record.message` not initialized, add tests

### DIFF
--- a/pytest_catchlog.py
+++ b/pytest_catchlog.py
@@ -126,13 +126,14 @@ class CatchLogPlugin(object):
     @contextmanager
     def _runtest_for(self, item, when):
         """Implements the internals of pytest_runtest_xxx() hook."""
-        with catching_logs(LogCaptureHandler(),
-                           formatter=self.formatter) as log_handler:
-            yield  # run test
+        with closing(py.io.TextIO()) as stream:
+            with catching_logs(logging.StreamHandler(stream),
+                               formatter=self.formatter):
+                yield  # run test
 
             if self.print_logs:
                 # Add a captured log section to the report.
-                log = log_handler.stream.getvalue().strip()
+                log = stream.getvalue().strip()
                 item.add_report_section(when, 'log', log)
 
     @pytest.mark.hookwrapper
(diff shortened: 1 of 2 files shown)


The-Compiler:
It seems this breaks using `LogRecord.message`:

```
    def test_abort_typeerror(question, qtbot, mocker, caplog):
        """Test Question.abort() with .emit() raising a TypeError."""
        signal_mock = mocker.patch('qutebrowser.utils.usertypes.Question.aborted')
        signal_mock.emit.side_effect = TypeError
        with caplog.at_level(logging.ERROR, 'misc'):
            question.abort()
>       assert caplog.records[0].message == 'Error while aborting question'
E       AttributeError: 'LogRecord' object has no attribute 'message'
```

It seems to work when I use the `caplog-rec-handler` branch from #24.

pytest_catchlog.py
```diff
@@ -268,6 +269,11 @@ class CallableStr(CallablePropertyMixin, str):
 class CompatLogCaptureFixture(LogCaptureFixture):
     """Backward compatibility with pytest-capturelog."""
 
+    def __init__(self, handler, item):
+        """Creates a new funcarg."""
```
The-Compiler, line 273: `funcarg` -> `fixture` :wink:

abusalimov, line 273: :+1:

abusalimov:
> It seems this breaks using `LogRecord.message`

Interesting enough. `LogRecord.message` is initialized by `Formatter.format()`, i.e. lazily. It is probably broken since replacing `StreamHandler` with simple `RecordingHandler`, which doesn't use formatting at all.

I'm not sure whether it is right to rely on that some Handler calls `format()`, but I guess, it would be better to add a workaround for this in pytest-catchlog.

I'll add a regression test for this, thank you! :+1:

PR #26, opened 2015-11-21
[RFC] Reorganize the source code into packages

abusalimov:
I'm working on pytest-catchlog performance optimization right now, and this likely will require some logging hacks that I would rather separate from the main plugin module.

@The-Compiler @eisensheng I'd like to hear you feedback and thoughts.

~~Note: this includes changes proposed in PR #24 and #25, **DO NOT MERGE** until these are accepted. It would be also easier to review, after that changes are merged into `develop`.~~


abusalimov:
Only 4 last commits matter, BTW:

  * f5cd65d `tests: Import pytest instead of py, use 'with pytest.raises():' syntax`
  * 1633c22 `tests: Use logging.getLogger(__name__) pattern, remove unused imports`
  * a8277da `tests: Move into tests/ and split into fixture and reporting tests`
  * 7b1a520 `Move under 'pytest_catchlog' package, extract fixture.py and common.py`

eisensheng:
> @The-Compiler @eisensheng I'd like to hear you feedback and thoughts.

I'm all in favor of this undertaking. I'll look at the changes tomorrow though, sorry.

The-Compiler:
I've only taken a quick look, but LGTM

PR #27, opened 2015-11-24
Update dependencies in tox.ini.

The-Compiler:
The listed pytest version doesn't work with Python 3.5.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,7 +3,7 @@ envlist = py{26,27,32,33,34,35}, pypy{,3}
 
 [testenv]
 deps =
-    py==1.4.28
-    pytest==2.7.1
+    py==1.4.30
+    pytest==2.8.3
 commands =
     py.test {posargs:test_pytest_catchlog.py}


The-Compiler:
I also suggest we set up [requires.io](https://requires.io/) for this project so we get notified when there's a new version - I don't think it can open pull requests for `tox.ini` unfortunately, but I can set it up to send a mail (to me, I think?) when dependencies change and then take care of that.

abusalimov:
I stumbled across this one recently too. What's about just using `>=` specifiers with the older version rather than sticking with `==` to a newer?

I mean changing it to `pytest>=2.7.1`.

The-Compiler:
I prefer pinning the version for reproducibility - otherwise you suddenly have some PR which seems like it'd break stuff, but actually it's because some new dependency was released and broke something.

Granted, I do this in a project which is [a bit bigger](https://github.com/The-Compiler/qutebrowser/blob/master/tox.ini), but I still think changing the pinned version manually once all few weeks (with requires.io) is better than having things break on random occasions (happened to me before I pinned versions, and not just once)

abusalimov:
Ah, I confused these with the requirements specified in `setup.py`, which set the minimal version. You're totally right, of course!

:shipit:

eisensheng:
> I also suggest we set up requires.io for this project so we get notified when there's a new version - I don't think it can open pull requests for tox.ini unfortunately, but I can set it up to send a mail (to me, I think?) when dependencies change and then take care of that.

Sounds like a good idea. Can't we set up a cheap mailing list for this plugin somewhere?

It might be a good idea to specify which pytest versions should be supported and test against all of them.

The-Compiler:
@eisensheng Sorry, I missed that message somehow

> Sounds like a good idea. Can't we set up a cheap mailing list for this plugin somewhere?

I can set up `pytest-catchlog@lists.schokokeks.org` as a mailinglist, but I think that's better deferred until after the rename.

> It might be a good idea to specify which pytest versions should be supported and test against all of them.

I'll try taking a look at this later this week.

eisensheng:
Anything that would prevent this from being merged now? We could work out support for multiple py.test versions later on in a separate issue and pull request. I'm stumbling about this now since I would like to kick out a new release and ran the test suite beforehand.

abusalimov:
> :shipit: 

LGTM

The-Compiler:
:+1:

eisensheng:
Thanks for looking over it again!

Issue #30, opened 2016-01-04
Unicode error in caplog.text


sirex:
`caplog.text` became a `CallableStr(CallablePropertyMixin, str)` property, witch always receive result from `py.io.TextIO.getvalue`, witch always returns `unicode`.

In other words, when you call `caplog.text` you get something like this: `str(unicode(''))`.

If logs have non-ascii characters it ends up like this:

``` pytb
  File "pytest_catchlog.py", line 250, in getter
    ret = cls(naked_value)
UnicodeEncodeError: 'ascii' codec can't encode character u'\u017d' in position 44726: ordinal not in range(128)
```

PR #17, opened 2015-11-10
Advertise support for Python 2.6 and test it in tox/travis

mnencia:
Add testing for python 2.6 compatibility in tox and travis.ci
Declare 2.6 support in setup.py metadata.

PR #31, opened 2016-01-04
Unicode support

sirex:
See https://github.com/eisensheng/pytest-catchlog/issues/30

Run these tests: `tox -e py26,py27,py34`, don't have other snakes.

### DIFF
--- a/pytest_catchlog.py
+++ b/pytest_catchlog.py
@@ -263,7 +263,7 @@ class CallablePropertyMixin(object):
 class CallableList(CallablePropertyMixin, list):
     pass
 
-class CallableStr(CallablePropertyMixin, str):
+class CallableStr(CallablePropertyMixin, py.builtin.text):
     pass
 
 
--- a/test_pytest_catchlog.py
+++ b/test_pytest_catchlog.py
@@ -1,3 +1,7 @@
+# coding: utf-8
+
+import sys
+
 import py
 
 pytest_plugins = 'pytester'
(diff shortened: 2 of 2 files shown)


The-Compiler:
You might want to use a backslash-escape (i.e. `\uXXXX`) instead of a literal to get things working with Python 3.2.

abusalimov:
Thanks for the contribution! Can we avoid the 'six' dependency? I'm pretty sure, there should be the same compat type somewhere in `py.builtin`.

The-Compiler:
`six.text_type` is `py.builtin.text` I think, and probably you can use `py.builtin._totext` instead of `six.u`.

sirex:
`py.builtin._totext` did not worked out of the box: https://bitbucket.org/pytest-dev/py/issues/75/pybuiltin_totext-does-not-work-with-python

But after replacing `_totext`, removed `six` dependency.

I think, `py` should use `six` too, instead of reinventing the wheel.

The-Compiler:
I bet `py` existed before `six` did (but I didn't check) :wink:

The long-term plan for pytest is to move away from pylib, though.

pytest_catchlog.py
```diff
@@ -311,3 +312,17 @@ def caplog(request):
...
+if sys.version_info >= (3, 0):
+    _totext = py.builtin._totext
+else:
+    def _totext(obj, encoding='utf-8', errors=None):
+        if isinstance(obj, str):
+            if errors is None:
+                obj = obj.decode(encoding)
+            else:
+                obj = obj.decode(encoding, errors)
+        return obj
```
abusalimov: It is only used in the test. Do we need it here?

sirex: I added it here to be able to access it from `testdir.makepyfile`. If there is a batter way how to make `_totext` visible from `testdir.makepyfile`, I can fix it.

abusalimov:
@The-Compiler maybe we could permit usage of `six` for testing, WDYT? Apparently, `_totext` is broken, and we need `u(...)` in tests...

The-Compiler:
I'd personally prefer using `u'...'` strings and dropping Python 3.2 support, like some other testing tools (like coverage.py or hypothesis) already did.

abusalimov:
@The-Compiler I agree, but let's wait at least until pytest-core drops 3.2 support first. After all, Python 2.6 support (which is about to be dropped in pytest 2.9) has been added just recently (#17). It would be unreasonable to abandon 3.2, but retain 2.6 compatibility.

Another option is to use `u'...'` literals and skip the test for Python 3.2. @sirex what do you think?

sirex:
To leave 3.2 support, while `pytest-core` still supports it seems reasonable to me.

But for this one unicode test I can add skip condition for 3.2, just to remove this `_totext` workaround.

abusalimov:
> just to remove this `_totext` workaround.

This is exactly what I meant :+1:

sirex:
Can't skip py32, because it fails to load whole file containing `u''`. So instead just added small `u()` lambdas that does just smallest minimum needed for tests to support all pythons.

abusalimov:
LGTM! Could you squash the whole branch into a single commit, please? :checkered_flag:

The-Compiler:
LGTM as well!

PR #33, opened 2016-01-25
[Simplified] Move pytest-logging functionality to pytest-catchlog

s0undt3ch:
This is a simplified and a more explicit approach to #20 

Closes #20.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -147,3 +147,26 @@ given severity and message::
             ('root', logging.INFO, 'boo arg'),
         ]
 
+
+Live Logs
+~~~~~~~~~
+
+By default, catchlog will output any logging records with a level higher or equal
+to WARNING. In order to actually see these logs in the console you have to disable
+pytest output capture by passing ``-s``.
+
+You can specify the logging level for which log records with equal or higher level
+are printed to the console by passing ``--log-cli-level``. This setting accepts the
+logging level names as seen in python's documentation or an integer as the logging
+level num.
+
+Additionally, you can also specify ``--log-cli-format`` and ``--log-cli-date-format``
+which mirror and default to ``--log-format`` and ``--log-date-format`` if not
+provided, but are applied only to the console logging handler.
+
+All of the CLI log options can also be set in the configuration INI file. The option
+names are:
+
+* ``log_cli_level``
+* ``log_cli_format``
+* ``log_cli_date_format``
--- a/pytest_catchlog/plugin.py
+++ b/pytest_catchlog/plugin.py
@@ -40,6 +40,12 @@ def pytest_addoption(parser):
         dest='log_print', action='store_const', const=False, default=True,
         help='disable printing caught logs on failed tests.'
     )
+    add_option_ini(
+        parser,
+        '--log-level',
+        dest='log_level', default=None,
+        help='logging level used by the logging module'
+    )
     add_option_ini(parser,
         '--log-format',
         dest='log_format', default=DEFAULT_LOG_FORMAT,
@@ -50,12 +56,58 @@ def pytest_addoption(parser):
         dest='log_date_format', default=DEFAULT_LOG_DATE_FORMAT,
         help='log date format as used by the logging module.'
     )
+    add_option_ini(
+        parser,
+        '--log-cli-level',
+        dest='log_cli_level', default=None,
+        help='cli logging level.'
+    )
+    add_option_ini(
+        parser,
+        '--log-cli-format',
+        dest='log_cli_format', default=None,
+        help='log format as used by the logging module.'
+    )
+    add_option_ini(
+        parser,
+        '--log-cli-date-format',
+        dest='log_cli_date_format', default=None,
+        help='log date format as used by the logging module.'
+    )
+
+
+def get_actual_log_level(config, setting_name):
+    """Return the actual logging level."""
+    log_level = get_option_ini(config, setting_name)
+    if not log_level:
+        return
+    if isinstance(log_level, py.builtin.text):
+        log_level = log_level.upper()
+    try:
+        return int(getattr(logging, log_level, log_level))
+    except ValueError:
+        # Python logging does not recognise this as a logging level
+        raise pytest.UsageError(
+            "'{0}' is not recognized as a logging level name for "
+            "'{1}'. Please consider passing the "
+            "logging level num instead.".format(
+                log_level,
+                setting_name))
 
 
 def pytest_configure(config):
     """Always register the log catcher plugin with py.test or tests can't
     find the  fixture function.
     """
+    log_cli_level = get_actual_log_level(config, 'log_cli_level')
+    if log_cli_level is None:
+        # No specific CLI logging level was provided, let's check
+        # log_level for a fallback
+        log_cli_level = get_actual_log_level(config, 'log_level')
+        if log_cli_level is None:
+            # No log_level was provided, default to WARNING
+            log_cli_level = logging.WARNING
+    config._catchlog_log_cli_level = log_cli_level
     config.pluginmanager.register(CatchLogPlugin(config), '_catch_log')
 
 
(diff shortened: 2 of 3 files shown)


README.rst
```diff
@@ -141,3 +141,27 @@ given severity and message::
...
         ]
 
+
+Live Logs
+~~~~~~~~~
+
+By default, catchlog will output any logging records with a level higher or equal
+to WARNING. In order to actually see these logs in the console you have to disable
+pytest output capture by passing ``-s``. The life span of the CLI logging handler
+lasts throughout the whole runtests session.
```
The-Compiler: nitpick: I'd just say "the whole test session" here, as someone not familiar with pytest internals might not know what you mean with "runtests session".

Alternatively maybe get rid of the whole sentence? This sounds like an implementation detail which is irrelevant to a catchlog user anyways, no?

README.rst
```diff
@@ -141,3 +141,27 @@ given severity and message::
...
+
+You can specify the logging level for which log records with equal or higher level
+are printed to the console by passing ``--log-cli-level``. This setting accepts the
+logging level names as seen in python's documentation or an integer as the logging
+level num.
+
+Additionally, you can also specify ``--log-cli-format`` and ``--log-cli-date-format``
+which are equal to ``--log-format`` and ``--log-date-format`` but are applied to the
```
abusalimov: Mb. replace _are equal to_ -> _correspond to_ or _mirror_?

abusalimov:
What's about adding a `--log-level` option, which would provide a fallback value for both `--log-cli-level` and `--log-file-level`? And making other options behave respectively, i.e. `--log-[date-]format` to be defaults for `--log-{cli,file}-[date-]format`.

README.rst
s0undt3ch: I'll get rid of the sentence.

README.rst
s0undt3ch: I'll switch it to mirror...

s0undt3ch:
I didn't initially add `--log-level` because that would suggest it was targeted to be used with the existing format options(which are to be used in tests, right)?

However, I see no harm, and its actually beneficial, to add the fallback support you suggest.

I'll ping you guys once I'm finished addressing your suggestions...

s0undt3ch:
What I mean with _it would suggest_ is that when users read the whole CLI help message and the available option args, they'd notice the CLI/file distinction and would come to that conclusion?

s0undt3ch:
PR updated.

abusalimov:
> existing format options(which are to be used in tests, right)?

That's true indeed, though #24 fixes that, i.e. tests will always use the standard formatting, regardless command line options.

s0undt3ch:
Ok! Cool!

s0undt3ch:
So, after #24 is merged, we remove the CLI specific options and use the original ones?

abusalimov:
I thought you propose CLI options to be used for live logs, don't you? In this sense, are file options _that_ different from CLI ones to remove the latter, but leave the former?

s0undt3ch:
Yes, CLI options are for live logs, however, I personally, in a CI context, want live logs less verbose and file logs more verbose, hence the separation.

abusalimov:
BTW, how is a log file supposed to be used on CI?

s0undt3ch:
We use Jenkins and we store it as an artifact for later inspection if need be.

s0undt3ch:
Is there anything that you'd like me to address on this PR?

abusalimov:
Honestly, I doubt the need to add a new set of command line options (`--log-cli-...`). Not that I'm against that, but adding an option makes the whole thing feel more complex, increases maintenance costs, etc. In other words, I would avoid that unless absolutely necessary.

Is it possible to get along with adding just the `--log-level` option and maybe reusing the existing `-s` (i.e. `--capture=no`)? @The-Compiler @eisensheng your thoughts?

s0undt3ch:
The problem is not the log level option, the problem is the format options which already exist and are not targeted towards CLI usage, they are targeted towards test usage.

When that stops being the case, after your other PR is merged, then we can drop the CLI options and reuse the existing ones.

So, either this goes in with the CLI options to avoid confusion, or, that other PR is merged first and the options are reused fotncli usage, since they won't be targeted for test usage anymore.

Did I make any sense?

abusalimov:
> Did I make any sense?

Yeah, totally.

I'll try to address remaining issues of that PR to get it merged then.

s0undt3ch:
Awesome!

s0undt3ch:
Rebased to resolve conflicts

s0undt3ch:
Where are we on this PR?

I think I need to rebase... Let me know of anything else required.

s0undt3ch:
I've solved conflicts again. Whats the status of this PR?

abusalimov:
I apologize this PR got stalled. The thing is, I have a new job, hence no spare time now, alas. @eisensheng @The-Compiler we need to get over with this feature, I feel ashamed =(

s0undt3ch:
No need to feel ashamed. I know about being busy, why do you think I haven't asked a month or two ago? :smile:
Of course I'd like to get this merged, my CI server is running off a branch on my fork with both of my PR's together, but as long as there's still will and you still find it useful, I'll keep updating it until it gets included in the project.

The-Compiler:
I agree we somehow need to get all the open stuff in (and then maybe talk about the rename to `pytest-logging` again) - I'm happy to contribute where I can, but I lack the time to take the lead here, sorry.

I've looked at this PR again, and it seems fine to me. I'm somewhat inclined to just merge it - is there anything still open holding this up?

s0undt3ch:
Nothing holding on my end and about the rename, just say when, although, I'd like to have my other PR merged in too before(which I'll have to rebase once this one is merged).

eisensheng:
I'm also very sorry regarding the current state of this issue. I planned to get back to this issue on this weekend but wasn't able to make it in the end unfortunately. 

What's left to do here for me is basically to review the code and the use cases over again, running the test suite, inspecting coverage and (hopefully) approving it. I scheduled the coming saturday for this.

s0undt3ch:
:+1:

pytest_catchlog/plugin.py
```diff
@@ -50,12 +56,60 @@ def pytest_addoption(parser):
...
+        '--log-cli-date-format',
+        dest='log_cli_date_format', default=None,
+        help='log date format as used by the logging module.'
+    )
+
+
+def get_actual_log_level(config, setting_name):
+    '''
+    Return the actual logging level
+    '''
```
eisensheng: Can you use double quotes for docstrings, please?

Please look at the bottom of https://www.python.org/dev/peps/pep-0257/#id15 for reference.

pytest_catchlog/plugin.py
```diff
@@ -50,12 +56,60 @@ def pytest_addoption(parser):
         dest='log_date_format', default=DEFAULT_LOG_DATE_FORMAT,
         help='log date format as used by the logging module.'
     )
+    add_option_ini(
```
eisensheng, line 59: It's not clear to me why there are two different options that appear to configure the same thing. Can you enlighten me here, please?

The-Compiler, line 59: There already was [some discussion](https://github.com/eisensheng/pytest-catchlog/pull/33#issuecomment-177315532) about this.

eisensheng, line 59: Duh! I looked through usage in source but missed that discussion. Sorry!

eisensheng:
@s0undt3ch What do you think about adding a test that actually tries to log a message and will appear on stderr accordingly? It might be a bit redundant but that test would be bound to the feature implemented here.

Otherwise it looks good to me apart from that one docstring nitpick and would like to merge it sooner than later.

I'm sorry if my tone was to harsh, it doesn't look to me like that but I was reading some (amusing) torvalds mails before reviewing the branch and I'm afraid his tone slipped in.

pytest_catchlog/plugin.py
s0undt3ch: I know that pep, i just choose not to follow for my own projects, I prefer single quotes.
I'll fix it in a bit.

s0undt3ch:
No worries about the tones, we all want the same goal, an excellent tool, and sometimes we just see some awkward code which brings out the tone(personal experience), or sometimes the goal is just not properly explained.

I'll add that test later in the evening or as soon as possible, I also want this in sooner than latter.

pytest_catchlog/plugin.py
The-Compiler: If we want to follow pep257, that should be `"""Return the actual logging level."""` - all in one line and with a period :wink:

s0undt3ch:
Single quote replaced with double quotes and tests updated to ensure we actually get the expected logging output in `stderr`

The-Compiler:
Wheee! :sparkles:

eisensheng:
This feature was merged on top of several more changes. It's difficult to draw a line there but the clean up of the entire plugin justifies the next version number to be 2.0.0 but there is still work left to do for the 2.0.0 release. Among that is the pytest-logging rename. This means it will take a little bit longer until this feature hits a published version. Is this okay for anyone interested in this feature or shall I do a 1.x release containing this feature beforehand?

s0undt3ch:
Woohooo!!!!

About waiting for 2.0, no worries on my end, I can keep installing from git.

PR #34, opened 2016-02-04
Allow configuring `log_print` in the ini file

s0undt3ch:


pytest_catchlog.py
```diff
@@ -71,15 +71,22 @@ def catching_logs(handler, filter=None, formatter=None,
...
+                                          help=help)
 
-def get_option_ini(config, name):
+
+def get_option_ini(config, name, default=None):
     ret = config.getoption(name)  # 'default' arg won't work as expected
     if ret is None:
         ret = config.getini(name)
+    if ret == '':
+        return default
```
abusalimov: I'm not quite sure about this change.

abusalimov:
LGTM apart from the nitpick above.

pytest_catchlog.py
s0undt3ch: Options not set in the ini file return an empty string... It's the only way we can trigger returning the default value. And we can't set the default value on the CLI option parser because then the ini option would never be checked against...

abusalimov: > we can't set the default value on the CLI option parser

But your change makes the opposite, doesn't it? I mean this diff (above):

``` diff
-    parser.getgroup('catchlog').addoption(option, dest=dest, help=help)
+    parser.getgroup('catchlog').addoption(option,
+                                          action=action,
+                                          dest=dest,
+                                          default=default,  # <- this line
+                                          help=help)
```

pytest_catchlog.py
```diff
@@ -71,15 +71,22 @@ def catching_logs(handler, filter=None, formatter=None,
...
 
-def add_option_ini(parser, option, dest, default=None, help=None):
+def add_option_ini(parser, option, dest, action=None, default=None, help=None):
     parser.addini(dest, default=default,
                   help='default value for ' + option)
-    parser.getgroup('catchlog').addoption(option, dest=dest, help=help)
+    parser.getgroup('catchlog').addoption(option,
+                                          action=action,
+                                          dest=dest,
+                                          default=default,
```
abusalimov: What's that for?

pytest_catchlog.py
abusalimov: If I'm not mistaken, `config.getini(name)` returns the default value, specified within `add_option_ini()`:

``` py
parser.addini(dest, default=default, ...)
```

pytest_catchlog.py
s0undt3ch: To pass the action to the option parser

abusalimov: Yeah, but what's about that particular line?

``` diff
+                                          default=default,
```

As you said,

> we can't set the default value on the CLI option parser because then the ini option would never be checked against...

pytest_catchlog.py
s0undt3ch: But we can't pass an option to the option parser, or else, `get_option_ini` would never get past `ret = config.getoption(name)` because the default value is set.

s0undt3ch: Also, if we don't pass the default as `None` to the option parser, the value of `ret = config.getoption(name)` will be `True` probably because the `action='store_false'` tell's it that the default should be `False` somewhere... And again, we wouldn't get past that value to investigate what the user has in the ini file.

pytest_catchlog.py
s0undt3ch: > Also, if we don't pass the default as None to the option parser, the value of `ret = config.getoption(name)` will be True probably because the `action='store_false'` tell's it that the default should be `False` somewhere... And again, we wouldn't get past that value to investigate what the user has in the ini file.

abusalimov: But that most likely would break the logic of setting the remaining options, like `log_format`, wouldn't that?

s0undt3ch: The tests pass... So I'm assuming it doesn't.

I think that on another PR of mine I've set default values directly when adding the option, and I probably shouldn't, defaults should be left at `None` and we pass the default value when getting the option with `get_option_ini` so both the CLI options and the INI options are looked at...

pytest_catchlog.py
abusalimov: Let's change it to `store_const`:

> `'store_const'` - This stores the value specified by the `const` keyword argument. (Note that the `const` keyword argument **defaults to the rather unhelpful `None`**.) The `'store_const'` action is most commonly used with optional arguments that specify some sort of flag.

I.e. _rather helpful_ in our case. :smile: 

> `'store_true'` and `'store_false'` - These are special cases of `'store_const'` using for storing the values `True` and `False` respectively. In addition, they create default values of `False` and `True` respectively.

https://docs.python.org/2/library/argparse.html#action

pytest_catchlog.py
abusalimov: > The tests pass... So I'm assuming it doesn't.

There is no such tests, unfortunately... :disappointed:

s0undt3ch:
Switched to `store_const`

s0undt3ch:
Updated

s0undt3ch:
INI setting mentioned it the README.

s0undt3ch:
Awesome!

README.rst
```diff
@@ -79,6 +79,12 @@ completely with::
 
     py.test --no-print-logs test_pytest_catchlog.py
 
+Or in you ``pytest.ini``::
```
s0undt3ch, line 82: Bah! Your, not you....

PR #39, opened 2016-03-18
Add option to set report section

lukas-bednar:
it may be helpful when you want to merge  logs with stdout/err.
or when you need to rename reporting section in general.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -93,6 +93,20 @@ Shows failed tests in the normal manner as no logs were captured::
     text going to stderr
     ==================== 2 failed in 0.02 seconds =====================
 
+
+You can change the name of section where is captured log reported to.
+You may find it useful when you want to merge captured logs with messages
+which were printed to stdout or stderr.
+It also may be useful when you want to change name of section in general.
+From CLI::
+
+    py.test --log-capture-report-section stdout
+
+Or in your ``pytest.ini``::
+
+    [pytest]
+    log_capture_report_section = stderr
+
 Inside tests it is possible to change the log level for the captured
 log messages.  This is supported by the ``caplog`` funcarg::
 
--- a/pytest_catchlog.py
+++ b/pytest_catchlog.py
@@ -101,6 +101,13 @@ def pytest_addoption(parser):
                    '--log-date-format',
                    dest='log_date_format', default=None,
                    help='log date format as used by the logging module.')
+    add_option_ini(parser,
+                   '--log-capture-report-section',
+                   dest='log_capture_report_section',
+                   default='log',
+                   help='Name of report section, default "log", you can also '
+                   'decide to report it as "stdout", or "stderr".')
+
 
 
 def pytest_configure(config):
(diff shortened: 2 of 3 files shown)


abusalimov:
I'm not sure there's a way to merge logs with stdout/stderr without a LogHandler redirecting logs to the appropriate stream. I like the idea in general though.

Could you make a test that (1) writes to stdout, then (2) logs, and then (3) writes to stdout _once again_. The reported output should appear interleaved, i.e. 1-2-3. I suspect it will show 1-3 in the first report section (stdout), and then 2 in the second _separate_ section (logs, that just happens to be also named `stdout`).

abusalimov:
Yep, it seems I was right, here's what I get:

```
   with: u'----------------------------- Captured stdout call -----------------------------'
nomatch: 'text going to stdout (1)\\n'
    and: u'text going to stdout (1)'
    and: u'text going to stdout (3)'
    and: u'----------------------------- Captured stdout call -----------------------------'
    and: u'test_change_report_section.py    6 INFO     log message going to stdout section (2)'
```

diff:

``` diff
diff --git a/test_pytest_catchlog.py b/test_pytest_catchlog.py
index 554988f..ed7c4da 100644
--- a/test_pytest_catchlog.py
+++ b/test_pytest_catchlog.py
@@ -340,15 +340,17 @@ def test_change_report_section(testdir):
         import logging

         def test_foo(caplog):
-            sys.stdout.write('text going to stdout')
-            logging.getLogger().info('log message going to stdout section')
+            sys.stdout.write('text going to stdout (1)\\n')
+            logging.getLogger().info('log message going to stdout section (2)')
+            sys.stdout.write('text going to stdout (3)\\n')
             assert False
         ''')
     result = testdir.runpytest()
     print(result.stdout)
     assert result.ret == 1
     result.stdout.fnmatch_lines(['*- Captured stdout call -*',
-                                 'text going to stdout',
-                                 '*log message going to stdout section'])
+                                 'text going to stdout (1)\\n',
+                                 '*log message going to stdout section (2)',
+                                 'text going to stdout (3)\\n'])
     py.test.raises(Exception, result.stdout.fnmatch_lines,
                    ['*- Captured *log call -*'])
```

lukas-bednar:
@abusalimov Yes, I am aware about that. Probably using 'merge' word wasn't good choice.

I can describe what problem I was trying to solve.
we use pytest-junitxml plugin to generate xml test report, this report also include captured stdout & stderr as `<system-out>` & `<system-err>` elements.
this report is collected by jenkins-junit plugin and visualized in Jenkins UI.

my problem is that pytest-junitxml plugin doesn't include 'log' section captured by this plugin into xml.
so then our QA/Dev department can not see these logs in jenkins report.

I was considering 3 ways how to make them visible in jenkins report.

1) patch pytest-junitxml plugin, I wanted to add option what report-section to include in result.
problem is that when I add `<system-log>` element, it fails on xsd validation.
so alternatively I could append it to <system-out> element, but I think this logic doesn't fit into this plugin. that led me to second option.

2) patch pytest-catchlog plugin to report logs in specific report-section, the thing was rename 'log' to 'stdout' / 'stderr' and it will get included in <system-out> element and then displayed in jenkins results.

3) go with 1) and create XSLT script which merge `<system-out>` with `<system-log>` together.

this patch is the implementation of option 2).
I hope I put some light on that, I am also open to any suggestion how to do it better.

lukas-bednar:
since it was merged to pytest core, it is not relevant. closing ...

PR #40, opened 2016-03-26
Performance test harness

abusalimov:
Measure an overhead of having the pytest-catchlog plugin activated, namely:
- `(caplog - stub)` the cost of requesting a `caplog` fixture (minus an overhead incurred by pytest to prepare an empty fixture)
- `log_emit` the cost of emitting a log message

These costs are evaluated in 4 different scenarios:
- default
- noplugin - when catchlog is not activated, i.e. running with `-pno:pytest_catchlog`
- noprint - when logs are not reported, i.e. running with `--no-print-logs`
- nocapture - when running with `-s`

Also add a cmdline option to plot multiple measurements for comparison. For example, to estimate how each commit in a branch affects performance and at the same time ensure branch bisectability before pushing it to submit a PR:

``` console
$ rm -rf ./.benchmarks  # clean up old measurements, if any
$ git rebase upstream/develop -i --exec 'tox -eperf'  # save and close an editor
...  # wait until done
$ tox -eperf -- --collect-only --perf-graph
```

A graph.svg file will be saved somewhere like ./.benchmarks/Linux-CPython-2.7-64bit/

@The-Compiler @eisensheng I'd like this to be merged ASAP at least since it doesn't affect any logic besides tests. Your feedback will be appreciated!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,20 @@ python:
   - "3.5"
   - "pypy"
   - "pypy3"
+before_install: |
+  pip install tox "virtualenv<14.0.0" # recent versions don't run on Python 3.2
+  if [ ! -v TOXENV ]
+  then
+    __py_ver=$TRAVIS_PYTHON_VERSION
+    export TOXENV=${__py_ver/[0-9].[0-9]/py${__py_ver/.}}
+  fi
+  function announce()
+    { echo -e "$ANSI_GREEN$@${ANSI_RESET}"; }
+  function tox()
+    { announce "Running tox in TOXENV=$TOXENV"; env tox "$@"; }
 
-install:
-  - pip install -e .
-script:
-  - py.test
+install: tox --notest
+script: tox
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,18 @@ python:
   - "3.5"
   - "pypy"
   - "pypy3"
+matrix:
+  include:
+    - python: "2.7"
+      env: TOXENV=perf
+
 before_install: |
   pip install tox "virtualenv<14.0.0" # recent versions don't run on Python 3.2
   if [ ! -v TOXENV ]
   then
     __py_ver=$TRAVIS_PYTHON_VERSION
-    export TOXENV=${__py_ver/[0-9].[0-9]/py${__py_ver/.}}
+    __tox_dfl=${__py_ver/[0-9].[0-9]/py${__py_ver/.}}
+    export TOXENV=${TOXENV:-$__tox_dfl}
   fi
   function announce()
     { echo -e "$ANSI_GREEN$@${ANSI_RESET}"; }
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,9 +30,11 @@ before_install: |
 install: tox --notest
 script: tox
 
+before_cache:
+  - rm -rf $HOME/.cache/pip/log
 cache:
   directories:
-    - $HOME/.cache/pip/http
+    - $HOME/.cache/pip
 
 deploy:
   provider: pypi
--- a/tests/perf/bench/test_runtest_hook.py
+++ b/tests/perf/bench/test_runtest_hook.py
@@ -1,0 +1,47 @@
+from __future__ import absolute_import, division, print_function
+
+import logging
+
+import pytest
+
+
+logger = logging.getLogger('pytest_catchlog.test.perf')
+
+
+@pytest.fixture(autouse=True)
+def bench_runtest(request, benchmark):
+    # Using benchmark.weave to patch a runtest hook doesn't seem to work with
+    # pytest 2.8.3; for some reason hook gets called more than once, before
+    # running benchmark cleanup finalizer, resulting in the
+    # "FixtureAlreadyUsed: Fixture can only be used once" error.
+    #
+    # Use plain old monkey patching instead.
+    ihook = request.node.ihook
+    saved_hook = ihook.pytest_runtest_call
+
+    def patched_hook(*args, **kwargs):
+        ihook.pytest_runtest_call = saved_hook  # restore early
+        return benchmark(saved_hook, *args, **kwargs)
+
+    ihook.pytest_runtest_call = patched_hook
+    benchmark.group = 'runtest'
+
+
+@pytest.yield_fixture  # because 'caplog' is also a yield_fixture
+def stub():
+    """No-op stub used in place of 'caplog'.
+
+    Helps to measure the inevitable overhead of the pytest fixture injector to
+    let us exclude it later on.
+    """
+    yield
+
+
+def test_fixture_stub(stub):
+    logger.info('Testing %r hook performance: %s',
+                'catchlog', 'pure runtest hookwrapper overhead')
+
+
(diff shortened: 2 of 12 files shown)


.travis.yml
```diff
@@ -9,15 +9,32 @@ python:
...
+  include:
+    - python: "2.7"
+      env: TOXENV=perf
 
-install:
-  - pip install -e .
-script:
-  - py.test
+before_install: |
+  pip install tox "virtualenv<14.0.0" # recent versions don't run on Python 3.2
```
The-Compiler, line 18: I'd rather drop 3.2 support already, as pip/virtualenv/coverage.py/hypothesis/... did as well. 

I know you argued against it in https://github.com/eisensheng/pytest-catchlog/pull/31#issuecomment-168746387 - but first 2.6 is less problematic, and I think it's in much more widespread use. Old distributions are going to be on 2.6/2.7, but if something is new enough to have Python 3, it'd at least have 3.3 (like Ubuntu Precise, the oldest still maintained Ubuntu LTS).

I'd rather avoid cumulating more and more technical debt for something probably not used by anyone.

abusalimov, line 18: Sounds reasonable to me. Although, I'd rather drop it in a separate PR, where we could get rid of a workaround from #31 too (I mean `u(...)` stuff).

BTW do you know, are there any plans to drop 3.2 in pytest-core?

The-Compiler, line 18: Okay, sounds good. pytest adopted the policy "we can drop it as soon as pip does", so we could drop it, but I don't think the need came up so far. Honestly I'm not sure how exactly pytest handles the text/unicode stuff so it's compatible with 2.6-3.5 without any major pains.

The-Compiler:
Crazy stuff you're doing there! :+1:

I only took a quick look, but LGTM - I'd like to see what @eisensheng thinks though :wink:

In the long run, some of this maybe could be upstreamed to pytest-benchmark? /cc @ionelmc

ionelmc:
Interesting, this looks like a tailored pytest-benchmark driver, including some custom plotting code. FYI I working to refactor some the internals in pytest-benchmark - I'd suspect you could at least use it's benchmark data selector in the future (it's glob-based tho).

tests/perf/conftest.py
```diff
@@ -0,0 +1,196 @@
...
+
+    terminalreporter.write_line(
+        'perf-graph: Saved graph into {0}'.format(output_file),
+        green=True, bold=True)
+
+
+@gen_dict
+def eval_benchmark_expr(expr, benchenvs, **kwargs):
+    for mode, envlist in benchenvs.items():
+        yield mode, [BenchmarkEnv(benchenv, **kwargs).evaluate(expr)
```
ionelmc, line 130: `evaluate` can return `None`. Shouldn't this filter those values?

abusalimov, line 130: Indeed it can, and this is intentional. Pygal takes care of `None` values properly by treating them as missing measurements data.

tests/perf/bench/test_runtest_hook.py
```diff
@@ -0,0 +1,47 @@
...
+    #
+    # Use plain old monkey patching instead.
+    ihook = request.node.ihook
+    saved_hook = ihook.pytest_runtest_call
+
+    def patched_hook(*args, **kwargs):
+        ihook.pytest_runtest_call = saved_hook  # restore early
+        return benchmark(saved_hook, *args, **kwargs)
+
+    ihook.pytest_runtest_call = patched_hook
```
abusalimov, line 26: @The-Compiler @ionelmc this is a workaround that I'm not confident about. Is there any pitfalls of doing that?

ionelmc, line 26: Do I understand this correctly - this will benchmark all the tests? Is this really necessary?

abusalimov, line 26: Yes, this will benchmark the tests defined in this file only (am I right? this won't apply to tests from other files, will it?).

Here we want to measure the cost of preparing our `caplog` fixture and then compare that with an overhead of having an empty stub fixture. Hence, two test cases, both being benchmarked.

ionelmc:
Another thing, is the idea of this benchmark driver to run all the tests with and without the benchmark?

abusalimov:
Well, there is a `--run-perf=check` option that runs performance tests with `--benchmark-disable` ([code](https://github.com/eisensheng/pytest-catchlog/pull/40/files#diff-9a3b6a42d3e9b1879cee16f75d55ddefR78)), though this is mostly for sanity checking reasons, i.e. to ensure these tests just compile, run and pass (this is what we also use for Travis builds).

The more general idea is to be able to:
- benchmark certain typical scenarios (see test cases in tests/perf/bench)
- measure how pytest invocation options affect that (plugin disabled / no reporting / no capturing, etc. - [code](https://github.com/eisensheng/pytest-catchlog/pull/40/files#diff-9a3b6a42d3e9b1879cee16f75d55ddef))
- and then plot everything to estimate dynamics from change to change.

And it is possible to plot not just benchmark results of a single test case, but also combine results of several test cases using an expression. For instance, by default `--perf-graph` would plot the results of `(caplog - stub)` expression, which is a shorthand for `(test_caplog_fixture - test_fixture_stub)`. That is, take the result of benchmarking `test_caplog_fixture` (the cost of preparing our fixture + inevitable overhead of the pytest itself) and subtract the result of `test_fixture_stub` (just the pytest overhead) to get the "pure" cost of making our `caplog` fixture.

abusalimov:
@The-Compiler thank you! :blush:

abusalimov:
@eisensheng Could you take a look once you have some spare time too, please?

tests/perf/test_perf_run.py
```diff
@@ -0,0 +1,87 @@
...
+    ]
+
+
+@pytest.fixture
+def bench_args(pytestconfig, storage_dir):
+    if pytestconfig.getoption('run_perf') != 'check':
+        return [
+            '--benchmark-only',
+            '--benchmark-disable-gc',
+            # '--benchmark-compare',
```
eisensheng: This comment doesn't look very meaningful to me. :innocent:

tests/perf/test_perf_run.py
```diff
@@ -0,0 +1,87 @@
...
+        return pytestconfig.option.color
+
+
+@pytest.fixture
+def verbosity(pytestconfig):
+    v = pytestconfig.option.verbose or -1
+    if v < 0:
+        return 'q' * (-v)
+    else:
+        return 'v' * v
```
eisensheng: What do you think about solving this with a ternary operator like in the `color` function? The variable `v` can be reused obviously.

tests/perf/test_perf_run.py
abusalimov: Nice catch! :+1:

tests/perf/test_perf_run.py
abusalimov: You mean like:

``` python
def verbosity(pytestconfig):
    v = pytestconfig.option.verbose or -1
    return ('v' if v > 0 else 'q') * abs(v)
```

eisensheng: Yep. :+1:

Issue #45, opened 2016-05-31
Provide reset method for records


blueyed:
As stated in https://github.com/eisensheng/pytest-catchlog/pull/21#issuecomment-156812911 there should be a `reset` method, especially since using `del caplog.records()[:]` will issue a deprecation warning.

PR #46, opened 2016-07-07
Add a clear method

blueyed:
This is required since `del caplog.records()[:]` will issue a
deprecation warning, because `records` became a property.

Fixes https://github.com/eisensheng/pytest-catchlog/issues/45.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -147,6 +147,13 @@ given severity and message::
             ('root', logging.INFO, 'boo arg'),
         ]
 
+You can call ``caplog.clear()`` to reset the captured log records in a test::
+
+    def test_something_with_clearing_records(caplog):
+        some_method_that_creates_log_records()
+        caplog.clear()
+        your_test_method()
+        assert ['Foo'] == [rec.message for rec in caplog.records]
 
 Live Logs
 ~~~~~~~~~
--- a/pytest_catchlog/fixture.py
+++ b/pytest_catchlog/fixture.py
@@ -42,6 +42,10 @@ class LogCaptureFixture(object):
         """
         return [(r.name, r.levelno, r.getMessage()) for r in self.records]
 
+    def clear(self):
+        """Reset the list of log records."""
+        self.handler.records = []
+
     def set_level(self, level, logger=None):
         """Sets the level for capturing of logs.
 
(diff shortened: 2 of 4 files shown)


abusalimov:
@blueyed Thank you! Could add a small description to the README, please?

Also I'm not sure about the best choice for naming the method. @eisensheng @The-Compiler what do you think? `reset_records()` / `reset()` / `flush()` / ...?

The-Compiler:
What about `clear()`? That's what Python uses for lists/dicts/sets/... so that's what I'd intuitively try.

blueyed:
I like `clear` - but it's not clear that it only clears/resets the records, is it?

The-Compiler:
I guess you could simply do `caplog.handler.records.clear()`? But that might be a bit too verbose, and you have a good point - `caplog.clear_records()` is what I'd prefer then.

abusalimov:
> that it only clears/resets the records

But what else could be cleared anyway?

The-Compiler:
> But what else could be cleared anyway?

I'm guessing `text` does not get cleared?

abusalimov:
But it does, in fact (implicitly).

The-Compiler:
Then, sure, `.clear()` :wink:

blueyed:
Awesome, amended and added to the README.

You did not ask for a test, should there be one?  (I guess you don't say "No"?!)

abusalimov:
> You did not ask for a test, should there be one? (I guess you don't say "No"?!)

Haha, you bet, I don't! :laughing: Tests would be appreciated, of course.

blueyed:
Digging my own grave, I'll add one later - but could take some days.

blueyed:
Finally came back to it: added tests, and refer to it from the deprecation warning, when `records` is accessed.

blueyed:
Fails on Python 3.2 (because of upgraded pytest).  See #49.

funkyfuture:
can i be of any help to get this tucked in?

abusalimov:
@blueyed Thank you for the efforts and sorry that it took so long to push that damn green button. This should have been merged long ago!

blueyed:
@abusalimov 
No worries, thanks for merging!

blueyed:
Anyone up for releasing it? :)

PR #61, opened 2017-04-01
Fix set_level side effects

### LINKED ISSUE #60: caplog.set_level has side effects if logger name is passed explicitly
If I pass logger name to caplog.set_level like this

```
caplog.set_level(logging.ERROR, logger='package.utils')
```

then for all following test cases that logger will have level set to ERROR, which is not expected.

chekunkov:
On test case completion all modifications to loggers levels are reversed.

Fixes #60

### DIFF
--- a/pytest_catchlog/fixture.py
+++ b/pytest_catchlog/fixture.py
@@ -13,9 +13,10 @@ from pytest_catchlog.common import catching_logs, logging_at_level
 class LogCaptureFixture(object):
     """Provides access and control of log capturing."""
 
-    def __init__(self, item):
+    def __init__(self, item, monkeypatch):
         """Creates a new funcarg."""
         self._item = item
+        self._monkeypatch = monkeypatch
 
     @property
     def handler(self):
@@ -55,7 +56,7 @@ class LogCaptureFixture(object):
         """
 
         obj = logger and logging.getLogger(logger) or self.handler
-        obj.setLevel(level)
+        self._monkeypatch.setattr(obj, 'level', logging._checkLevel(level))
 
     def at_level(self, level, logger=None):
         """Context manager that sets the level for capturing of logs.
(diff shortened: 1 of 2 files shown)


chekunkov:
Have no idea why pypy3 tests fail, they are broken in master. 

Regards 2.6 tests failure - I have two options:

1. Have if statement with some fix only for Py 2.6.
2. Drop .Py 2.6 support.

I'd go with the latter option, it seems to be trending among major Python packages:

https://github.com/pypa/pip/issues/3955
https://github.com/pypa/setuptools/issues/878 
https://github.com/pytest-dev/pytest/pull/2031

WDYT?

hugovk:
It's time to drop 2.6, it's been EOL since 2013:

* https://en.wikipedia.org/wiki/CPython#Version_history
* https://snarky.ca/stop-using-python-2-6/
* http://www.curiousefficiency.org/posts/2015/04/stop-supporting-python26.html
* http://www.python3statement.org
* Current pip 9 deprecates Python 2.6 support, pip 10 won't support it (https://github.com/pypa/pip/issues/3955)
* Not much PyPI traffic (June 2016) https://github.com/pypa/pip/issues/3796

hugovk:
Here's the pip installs for pytest-catchlog from PyPI for the last month (via `pypinfo --percent --pip pytest-catchlog pyversion`)

| python_version | percent | download_count |
| -------------- | -------: | --------------: |
| 2.7            |   57.1% |         25,939 |
| 3.6            |   18.4% |          8,337 |
| 3.5            |   14.4% |          6,535 |
| 3.4            |    6.6% |          3,014 |
| 3.3            |    2.6% |          1,194 |
| 3.7            |    0.8% |            363 |
| 2.6            |    0.1% |             25 |

I think it's safe to drop 2.6!

PR #70, opened 2017-07-24
Add caplog.grep(regex, level, name) and tests for it.

the9000:
* .grep has an extensive doc comment, very compact code.
* The tests seem exhaustive.
* All tests pass.


the9000:
The invalid pypy3 tox config will be fixed separately.

The-Compiler (approved):
This seems really useful!

the9000:
It's way obsolete.

